PR #17, opened 2022-04-19
Fixes for a working set of components

adamdbrw:
Changes required for Components to work correctly:
- display in search
- have editable fields

Static transform broadcaster needs to be a singleton - it has been moved to system component. However, TransformPublisher remains and will be responsible for dynamic transforms (which will reuse the same message creation code).


adamdbrw:
Incorporated in #18, closing

Issue #12, opened 2022-04-19
Publish dynamic transforms


adamdbrw:
Dynamic transforms are used to inform robotic system about relationship between two or more frames which change position and orientation in relationship to each other. For example, a moving robot will move it's base frame (base_link) in relationship to global frame ("world" or "map). Similarly, a mobile robotic arm will have a dynamic transform in relationship to robot / arm base.

Currently, only static transforms are published. Unlike static transforms, dynamic transforms need to be published each frame (each time they change). Also, they use a different topic "tf" (as opposed to "tf_static") and different publisher entity. The behavior can be captured in TransformPublisher, which already contains logic to compute the transformation, but only does so once and for static transform broadcaster.

The task is to extend this logic and also add a boolean m_publishTransform field to ROS2FrameComponent to control whether transforms where this frame is a child frame are published.

adamdbrw:
Fixed by #21

PR #18, opened 2022-04-19
Sensor abstraction improvements

adamdbrw:
Fixes and improvements to sensor abstraction.
Includes dynamic transform publishing.
Implements #12 (but to fix issue, "main" branch merge is required).


Code/Source/Transform/ROS2FrameComponent.cpp
```diff
@@ -6,20 +6,57 @@
...
+            {
+                AZ::TickBus::Handler::BusDisconnect();
+            }
+            m_ros2Transform.reset();
+        }
+    }
+
+    void ROS2FrameComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
+    {
+        m_ros2Transform->Publish(GetFrameTransform());
```
pijaro, line 48: I think we should be able to control the working frequency of the non-dynamic transform publisher. Maybe add TODO or a note that it is locked to the FPS of the simulator for the current implementation.

pijaro (approved):

Issue #10, opened 2022-04-19
Implement handling of namespaces


adamdbrw:
Currently, namespaces are handled in a very basic way, where user can input namespace to any ROS2 Frame Component. However, they need to meet the following requirements:
- [x] Hierarchical composition (compose also using parent frames for nested namespaces)
- [x] Select either "empty", "generated from entity name" or "custom". Implement Editor behavior and logic accordingly
- [x] "empty" by default, except top-level frame (Frame Component with no parent), which is "generated from entity name"
- [x] Implement generation from entity name, mining adherence to ros2 naming principles (see #7)
- [x] namespace edit text field is only visible to edit when "custom" is selected

adamdbrw:
Closed by #21

PR #19, opened 2022-04-20
Namespaces

adamdbrw:
Develops #10


adamdbrw:
Note that some of the more complex interactions (e.g. showing combined value or value based on component hierarchy) require a dedicated EditorComponent, which I believe can be postponed until after milestone 1

Code/Source/Frame/NamespaceConfiguration.h
```diff
@@ -0,0 +1,49 @@
...
+
+        void PopulateNamespace(bool isTop, const AZStd::string &entityName);
+        AZStd::string GetNamespace(const AZStd::string& parentNamespace) const;
+
+        static void Reflect(AZ::ReflectContext* context);
+
+    private:
+        AZStd::string m_namespace;
+        NamespaceStrategy m_namespaceStrategy;
+        bool m_isTop;
```
pijaro, line 41: Little confused by this variable name. Maybe `m_isRoot` would fit it better?

Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -54,9 +55,14 @@ namespace ROS2
...
 
+    bool ROS2FrameComponent::IsTopLevel() const
+    {
+        return GetGlobalFrameName() == GetParentFrameID();
+    }
+
     bool ROS2FrameComponent::IsDynamic() const
     {   // TODO - determine by joint type
-        return GetGlobalFrameName() == GetParentFrameID();
+        return IsTopLevel();
```
pijaro, line 65: Not a fan of this assumption. Stationary robots also exist (i.e. the root of a stationary manipulator - base-link - in a warehouse map).

adamdbrw: Sure, but this will be determined based on joint type as in TODO - and also perhaps on another type of description. Not for now, for later

Code/Source/Frame/NamespaceConfiguration.h
adamdbrw: isTopLevel? Or isRoot?

PR #20, opened 2022-04-20
Initial pr

adamdbrw:
Re-cast PR after force push due to signing of first commit on main


adamdbrw:
Same content as #20, fixed unsigned commits (2) including one in main. 
PR went through review already.

Issue #16, opened 2022-04-19
Add QoS to sensor configuration


adamdbrw:
Add QoS to sensor configuration, at least support these values:
- [x] Best Effort / Reliable
- [x] Keep Last Depth History

Default to Best Effort, Keep Last 10. Please note that QoS should be a part of Publisher abstraction that is due to emerge from #14

PR #21, opened 2022-04-21
ROS2Frame component:

### LINKED ISSUE #10: Implement handling of namespaces

### LINKED ISSUE #12: Publish dynamic transforms

adamdbrw:
- Static frames publishing (e.g. for fixed joints, many sensors)
- Dynamic frame publishing (to global frame) - Fixes #12
- Handling of namespaces - Fixes #10
- Several utilities for ROS2 conversions and names needed for frames/transforms

### DIFF
--- a/Code/Source/Frame/NamespaceConfiguration.cpp
+++ b/Code/Source/Frame/NamespaceConfiguration.cpp
@@ -61,7 +61,7 @@ namespace ROS2
         return ROS2Names::GetNamespacedName(parentNamespace, m_namespace);
     }
 
-    bool NamespaceConfiguration::IsNamespaceCustom()
+    bool NamespaceConfiguration::IsNamespaceCustom() const
     {
         return m_namespaceStrategy == NamespaceConfiguration::NamespaceStrategy::Custom;
     }
@@ -80,7 +80,7 @@ namespace ROS2
             {
                 ec->Class<NamespaceConfiguration>("Namespace Configuration", "Handles ROS2 namespaces")
                     ->DataElement(AZ::Edit::UIHandlers::ComboBox, &NamespaceConfiguration::m_namespaceStrategy,
-                            "Namespace strategy", "Determines how namespace for frames and topics is created in hierarchy")
+                            "Namespace strategy", "Determines how namespace for frames and topics is created from the hierarchy")
                         ->Attribute(AZ::Edit::Attributes::ChangeNotify, &NamespaceConfiguration::OnNamespaceStrategySelected)
                         ->EnumAttribute(NamespaceConfiguration::NamespaceStrategy::Default, "Default")
                         ->EnumAttribute(NamespaceConfiguration::NamespaceStrategy::Empty, "Empty")
--- a/Code/Source/Frame/NamespaceConfiguration.cpp
+++ b/Code/Source/Frame/NamespaceConfiguration.cpp
@@ -24,19 +24,23 @@ namespace ROS2
 
     void NamespaceConfiguration::UpdateNamespace()
     {
-        auto nss = m_namespaceStrategy;
-        if (NamespaceStrategy::Custom == nss)
-        {   // Leave the name as it is
-            return;
-        }
-
-        if (NamespaceStrategy::Empty == nss || (!m_isRoot && NamespaceStrategy::Default == nss))
+        switch(m_namespaceStrategy)
         {
-            m_namespace = "";
-        }
-        else if (NamespaceStrategy::FromEntityName == nss || (m_isRoot && NamespaceStrategy::Default == nss))
-        {
-            m_namespace = ROS2Names::RosifyName(m_entityName);
+            case NamespaceStrategy::Empty:
+                m_namespace = "";
+                break;
+            case NamespaceStrategy::Default:
+                m_namespace = m_isRoot ? ROS2Names::RosifyName(m_entityName) : "";
+                break;
+            case NamespaceStrategy::FromEntityName:
+                m_namespace = ROS2Names::RosifyName(m_entityName);
+                break;
+            case NamespaceStrategy::Custom:
+                // Leave the namespace as is
+                break;
+            default:
+                AZ_Assert(false, "Unhandled namespace strategy");
+                break;
         }
     }
 
--- a/Code/Source/Frame/NamespaceConfiguration.cpp
+++ b/Code/Source/Frame/NamespaceConfiguration.cpp
@@ -15,10 +15,10 @@
 
 namespace ROS2
 {
-    void NamespaceConfiguration::PopulateNamespace(bool isRoot, const AZStd::string &entityName)
+    void NamespaceConfiguration::PopulateNamespace(bool isRoot, AZStd::string entityName)
     {
         m_isRoot = isRoot;
-        m_entityName = entityName;
+        m_entityName = AZStd::move(entityName);
         OnNamespaceStrategySelected();
     }
 
--- a/Code/Source/Frame/NamespaceConfiguration.cpp
+++ b/Code/Source/Frame/NamespaceConfiguration.cpp
@@ -1,0 +1,96 @@
+/*
+* Copyright (c) Contributors to the Open 3D Engine Project.
+* For complete copyright and license terms please see the LICENSE at the root of this distribution.
+*
+* SPDX-License-Identifier: Apache-2.0 OR MIT
+*
+*/
+
+#pragma once
+
+#include "Frame/NamespaceConfiguration.h"
+#include "Utilities/ROS2Names.h"
+#include <AzCore/Serialization/EditContext.h>
+#include <AzCore/Serialization/EditContextConstants.inl>
+
+namespace ROS2
+{
+    void NamespaceConfiguration::PopulateNamespace(bool isRoot, const AZStd::string &entityName)
+    {
+        m_isRoot = isRoot;
+        m_entityName = entityName;
+        OnNamespaceStrategySelected();
+    }
+
+    void NamespaceConfiguration::UpdateNamespace()
+    {
+        auto nss = m_namespaceStrategy;
+        if (NamespaceStrategy::Custom == nss)
+        {   // Leave the name as it is
+            return;
+        }
+
+        if (NamespaceStrategy::Empty == nss || (!m_isRoot && NamespaceStrategy::Default == nss))
+        {
+            m_namespace = "";
+        }
+        else if (NamespaceStrategy::FromEntityName == nss || (m_isRoot && NamespaceStrategy::Default == nss))
+        {
+            m_namespace = ROS2Names::RosifyName(m_entityName);
+        }
+    }
+
+    AZ::Crc32 NamespaceConfiguration::OnNamespaceStrategySelected()
+    {
+        UpdateNamespace();
+        return AZ::Edit::PropertyRefreshLevels::EntireTree;
+    }
+
+    AZStd::string NamespaceConfiguration::GetNamespace(const AZStd::string& parentNamespace) const
+    {
+        if (parentNamespace.empty())
+        {
+            return m_namespace;
+        }
+
+        if (m_namespace.empty())
+        {
+            return parentNamespace;
+        }
+
+        return ROS2Names::GetNamespacedName(parentNamespace, m_namespace);
+    }
+
+    bool NamespaceConfiguration::IsNamespaceCustom()
+    {
+        return m_namespaceStrategy == NamespaceConfiguration::NamespaceStrategy::Custom;
+    }
+
+    void NamespaceConfiguration::Reflect(AZ::ReflectContext* context)
+    {
+        if (auto serializeContext = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serializeContext->Class<NamespaceConfiguration>()
+                ->Version(1)
+                ->Field("Namespace Strategy", &NamespaceConfiguration::m_namespaceStrategy)
+                ->Field("Namespace", &NamespaceConfiguration::m_namespace)
+                ;
+
+            if (AZ::EditContext* ec = serializeContext->GetEditContext())
+            {
+                ec->Class<NamespaceConfiguration>("Namespace Configuration", "Handles ROS2 namespaces")
+                    ->DataElement(AZ::Edit::UIHandlers::ComboBox, &NamespaceConfiguration::m_namespaceStrategy,
+                            "Namespace strategy", "Determines how namespace for frames and topics is created in hierarchy")
+                        ->Attribute(AZ::Edit::Attributes::ChangeNotify, &NamespaceConfiguration::OnNamespaceStrategySelected)
+                        ->EnumAttribute(NamespaceConfiguration::NamespaceStrategy::Default, "Default")
+                        ->EnumAttribute(NamespaceConfiguration::NamespaceStrategy::Empty, "Empty")
+                        ->EnumAttribute(NamespaceConfiguration::NamespaceStrategy::FromEntityName, "Generate from entity name")
+                        ->EnumAttribute(NamespaceConfiguration::NamespaceStrategy::Custom, "Custom")
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &NamespaceConfiguration::m_namespace, "Namespace", "Namespace")
+                        ->Attribute(AZ::Edit::Attributes::Visibility, &NamespaceConfiguration::IsNamespaceCustom)
+                        // TODO - hide for now, but desired Editor component behavior would be to show a read only value
+                        ;
+            }
+        }
+    }
+}  // namespace ROS2
--- a/Code/Source/Frame/NamespaceConfiguration.h
+++ b/Code/Source/Frame/NamespaceConfiguration.h
@@ -37,11 +37,11 @@ namespace ROS2
 
     private:
         AZStd::string m_namespace;
-        NamespaceStrategy m_namespaceStrategy;
+        NamespaceStrategy m_namespaceStrategy = NamespaceStrategy::Default;
         bool m_isRoot;
         AZStd::string m_entityName;
 
-        bool IsNamespaceCustom();
+        bool IsNamespaceCustom() const;
         void UpdateNamespace();
         AZ::Crc32 OnNamespaceStrategySelected();
     };
--- a/Code/Source/Frame/NamespaceConfiguration.h
+++ b/Code/Source/Frame/NamespaceConfiguration.h
@@ -8,19 +8,20 @@
 
 #pragma once
 
-#include <AzCore/std/string/string.h>
 #include <AzCore/RTTI/RTTI.h>
 #include <AzCore/Serialization/SerializeContext.h>
+#include <AzCore/std/string/string.h>
 
 namespace ROS2
 {
-    /// Configuration for handling of namespaces
+    /// Configuration for handling of namespaces. Namespaces are useful for various ROS 2 components.
     struct NamespaceConfiguration
     {
     public:
         AZ_RTTI(NamespaceConfiguration, "{5E5BC6EA-DD01-480E-A4D1-6857CF70FDC8}");
         NamespaceConfiguration() = default;
         virtual ~NamespaceConfiguration() = default;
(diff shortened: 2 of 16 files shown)


Code/Source/Frame/NamespaceConfiguration.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <AzCore/Serialization/EditContextConstants.inl>
+
+namespace ROS2
+{
+    void NamespaceConfiguration::PopulateNamespace(bool isRoot, const AZStd::string &entityName)
+    {
+        m_isRoot = isRoot;
+        m_entityName = entityName;
+        OnNamespaceStrategySelected();
+    }
```
lumberyard-employee-dm, line 23: FYI if you accept the `entityName` parameter by value, you can use `AZStd::move` to perform a move operation to store it in m_entityName, instead of making a copy of it.
```suggestion
    void NamespaceConfiguration::PopulateNamespace(bool isRoot, const AZStd::string &entityName)
    {
        m_isRoot = isRoot;
        m_entityName = AZStd::move(entityName);
        OnNamespaceStrategySelected();
    }
```

Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -0,0 +1,158 @@
...
+    void ROS2FrameComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
+    {
+        m_ros2Transform->Publish(GetFrameTransform());
+    }
+
+    AZStd::string ROS2FrameComponent::GetGlobalFrameName()
+    {
+        // TODO - parametrize this, sometimes it is "world" and sometimes "map".
+        return "world";
+    }
```
lumberyard-employee-dm, line 57: If you update the return value to be a `const char*`, it would provide a bit more flexibility for the caller.

That would allow the callers to store the value in a `std::string` if they chose.

Code/Source/Frame/ROS2FrameComponent.h
```diff
@@ -0,0 +1,69 @@
...
+    /// This component marks an interesting reference frame and captures common behavior
+    /// for sensor data frame of reference, publishing ros2 static and dynamic transforms (tf)
+    /// and helping to export robot prefab to URDF. A robot should have a ROS2Frame defined for every
+    /// entity (each level).
+    class ROS2FrameComponent
+        : public AZ::Component
+        , public AZ::TickBus::Handler
+    {
+    public:
+        AZ_COMPONENT(ROS2FrameComponent, "{EE743472-3E25-41EA-961B-14096AC1D66F}", AZ::Component);
```
lumberyard-employee-dm, line 27: The `AZ::Component` base class doesn't need to be specified when using the `AZ_COMPONENT` macro.
That macro already adds [AZ::Component](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Component/Component.h#L312) on behalf of the caller

Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -0,0 +1,88 @@
...
+        {
+        public:
+            void Publish(const geometry_msgs::msg::TransformStamped& transformMessage) override
+            {
+                if (m_published)
+                    return; // Consider adding a warning here as well
+
+                return ROS2Interface::Get()->BroadcastStaticTransform(transformMessage);
+                m_published = true;
+            }
```
lumberyard-employee-dm, line 27: This code will warn on clang and and because O3DE compilers with warrnings as errors on, this will trigger an error when using this gem in O3DE with the clang compiler.

The `m_published = true` needs to be before the return statement.

lumberyard-employee-dm (changes requested):

Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        AZStd::unique_ptr<TransformPublisher> TransformPublisher::CreateTransformPublisher(bool isDynamic)
+        {
+            if (isDynamic)
+            {
+                return std::make_unique<DynamicTransformPublisher>();
+            }
+            else
+            {
+                return std::make_unique<StaticTransformPublisher>();
```
lumberyard-employee-dm, line 62: AZStd::make_unique should be used here as there is a [bit of logic](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/smart_ptr/unique_ptr.h#L41-L61) to determine whether `aznew` should be used to use the AZ Allocators base on the AZ_CLASS_ALLOCATOR macro

Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+    }
+
+    ROS2Transform::ROS2Transform(const AZStd::string& parentFrame, const AZStd::string& childFrame, bool isDynamic)
+        : m_parentFrame(parentFrame), m_childFrame(childFrame)
```
lumberyard-employee-dm, line 68: optimization:" You could accept parentFrame and childFrame by value and then move them into the `m_parentFrame` and `m_childFrame` member variables.

Code/Source/Frame/NamespaceConfiguration.cpp
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if (NamespaceStrategy::Empty == nss || (!m_isRoot && NamespaceStrategy::Default == nss))
+        {
+            m_namespace = "";
+        }
+        else if (NamespaceStrategy::FromEntityName == nss || (m_isRoot && NamespaceStrategy::Default == nss))
+        {
+            m_namespace = ROS2Names::RosifyName(m_entityName);
+        }
```
spham-amzn, line 40: Suggestion: For readability, can we 'switch' this to a switch statement?
```suggestion
        switch(m_namespaceStrategy)
        {
            case NamespaceStrategy::Empty:
                m_namespace = "";
                break;
            case NamespaceStrategy::Default:
                m_namespace = ROS2Names::RosifyName(m_entityName) ? m_isRoot : "";
                break;
            case NamespaceStrategy::FromEntity:
                m_namespace = ROS2Names::RosifyName(m_entityName);
                break;
            case NamespaceStrategy::Custom:
                // Leave the namespace as is
                break;
             default:
                AZ_Assert(false, "Unhandled namespace strategy");
                break;
        }
```

Code/Source/Frame/NamespaceConfiguration.cpp
```diff
@@ -0,0 +1,96 @@
...
+                ->Version(1)
+                ->Field("Namespace Strategy", &NamespaceConfiguration::m_namespaceStrategy)
+                ->Field("Namespace", &NamespaceConfiguration::m_namespace)
+                ;
+
+            if (AZ::EditContext* ec = serializeContext->GetEditContext())
+            {
+                ec->Class<NamespaceConfiguration>("Namespace Configuration", "Handles ROS2 namespaces")
+                    ->DataElement(AZ::Edit::UIHandlers::ComboBox, &NamespaceConfiguration::m_namespaceStrategy,
+                            "Namespace strategy", "Determines how namespace for frames and topics is created in hierarchy")
```
spham-amzn, line 83: should this read 
```
Determines how namespace for frames and topics is created from the hierarchy
```
?

Code/Source/Frame/NamespaceConfiguration.h
```diff
@@ -0,0 +1,49 @@
...
+        void PopulateNamespace(bool isRoot, const AZStd::string &entityName);
+        AZStd::string GetNamespace(const AZStd::string& parentNamespace) const;
+
+    private:
+        AZStd::string m_namespace;
+        NamespaceStrategy m_namespaceStrategy;
+        bool m_isRoot;
+        AZStd::string m_entityName;
+
+        bool IsNamespaceCustom();
```
spham-amzn, line 44: minor:
could make this a const
```suggestion
        bool IsNamespaceCustom() const;
```

Code/Source/Frame/NamespaceConfiguration.cpp
spham-amzn: The suggestion above needs to replace ```const AZStd::string &``` with ```AZStd::string``` then

Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        class DynamicTransformPublisher : public TransformPublisher
+        {
+        public:
+            DynamicTransformPublisher()
+            {
+                auto ros2Node = ROS2Interface::Get()->GetNode();
+
+                // TODO - determine whether default QoS is good and what is good practice for queue size. Compare to tf2 broadcaster
+                const size_t keepLastQueueSize = 10;
```
spham-amzn, line 41: not for this review, but in the future, if we want to measure the impacts of different queue values, this value can be read from the settings registry instead

adamdbrw: Typically, ros2 developer sets the queue size as part of Quality of Service settings for each publisher, subscribers. There are default profiles to use (e.g. System, Sensor, Parameters), but most users are used to specifying queue size (History depth) for pubs/subs. This task will monitor the work on QoS: https://github.com/RobotecAI/o3de-ros2-gem/issues/16 for sensors. For transforms, there is likely a standard practice or some considerations that I need to check and set the value accordingly.

Code/Source/Frame/ROS2Transform.cpp
adamdbrw: Good catch. I am compiling with clang and somehow I did not get any warning. How is this possible - I will look into it.

Code/Source/Frame/ROS2Transform.cpp
adamdbrw: Could you explain reasoning behind this suggestion?
I see two possible advantages for your suggestion here:
1) would make no copy for rvalue arguments (2 moves?)
2) would make it explicit that function makes a copy of frame arguments (in signature, without needing to look elsewhere).

Anyway, I would like to understand your reasoning since I am not used to optimizing functions that are called this rarely.

Code/Source/Frame/NamespaceConfiguration.cpp
adamdbrw: I decided against a switch here since I thought there will be no more strategies to add and one early return + one if/else looks better. But I can change it, no strong opinion here, just a bit of dislike for switch.

Code/Source/Frame/NamespaceConfiguration.cpp
adamdbrw: wouldn't that result in unnecessary deallocation of m_entityName? A copy would reuse the storage

Code/Source/Frame/ROS2Transform.cpp
adamdbrw: I got something better - right from the ros2 tf2_ros library as defined there: https://docs.ros.org/en/ros2_packages/rolling/api/tf2_ros/generated/classtf2__ros_1_1DynamicBroadcasterQoS.html#exhale-class-classtf2-ros-1-1dynamicbroadcasterqos

Code/Source/Frame/NamespaceConfiguration.cpp
adamdbrw: I went with it, only switched: `m_namespace = ROS2Names::RosifyName(m_entityName) ? m_isRoot : "";`
to `m_namespace = m_isRoot ? ROS2Names::RosifyName(m_entityName) : "";`
and FromEntity -> FromEntityName

adamdbrw:
I resolved most of the review comments, but with two of them I would like to have a better understanding of the benefits before I apply them. I am aware of the general practice for function signatures and move semantics, but in these cases I would also like to perhaps learn a bit from your feedback.

Note that commits with review changes were first applied to development and then cherry-picked into ros2_frame (this PR). This is simply to make my life easier when preparing subsequent PRs.

Code/Source/Frame/ROS2Transform.cpp
lumberyard-employee-dm: The reason for my suggestion is based on how this function could be called.

### Scenario 1 - const char(&)[N] string literal
If the user constructs `ROS2Transform` with string literal parameters for the `parentFrame` and `childFrame` than a `AZStd::string` temporary instance has to be created and abound to the `const AZStd::string&` type.
Afterwards the copy constructor of the AZStd::string class is then invoked to make a copy of the temporary instance.

```
ROS2Transform transform1("Parent", "Child", true);
// before the call - Two AZStd::strings are constructed and bound to the ROS2Transform constructor parameters 
// during the call - Two copy constructor calls are invoked to store the m_parentFrame and m_childFrame members
```

With the signature of `const AZStd::string` signature, the AZStd::string `const char*` constructor has to be invoked 2 times + the AZStd::string copy constructor has to be invoked 2 times as well
So total operations = 4 copies

Using a signature of `AZStd::string`, the AZStd::string `const char*` constructor has to be invoked 2 times + the AZStd::string move constructor has to be invoked 2 times as well.
So total operations = 2 copies + 2 moves

In this scenario the `AZStd::string` signature wins

### Scenario 1 - AZStd::string copy
Next let's move on to the situation where the constructor is called with an `AZStd::string` call where a copy is wanted.
```
AZStd::string memberParentFrame = "MemberParent";
AZStd::string memberChildFrame = "MemberChild";
// ...
ROS2Transform transform2(memberParentFrame, memberChildFrame, true);
```

With the signature of `const AZStd::string&`, the `memberParentFrame` and `memberChildFrame` variables are bound to reference, so that operation is free, but because the parameter is `const AZStd::string&`, the ROS2Transform constructor has to make a copy
So total operations = 2 copies

With the signature of `AZStd::string`, the `memberParentFrame` and `memberChildFrame` variables values would have to  invoke the copy constructor to bind it to the `parentFrame` and `childFrame` parameters, but inside of the ROS2Transform constructor only a move constructor operation has to be performed.
So total operations  = 2 copies + 2 moves

In this scenario `const AZStd::string&` signature wins

### Scenario 3 - AZStd::string move
In this situation there is an AZStd::string variable on the stack, but it is moved during the call to the ROS2Transform constructor
```
AZStd::string transientParentFrame = "TransientParent";
AZStd::string transientChildFrame = "TransientChild";
// ...
ROS2Transform transform3(AZStd::move(transientParentFrame), AZStd::move(transientChildFrame), true);
```

Now if the ROS2Transform constructor had the signature of  `const AZStd::string&`, the `AZStd::move(transientParentFrame)` and `AZStd::move(transientChildFrame)` expression would cast to the type of `AZStd::string&&` which due to reference binding rules is allowed to bind to `const AZStd::string&`.
Now because the parameter signature is still `const AZStd::string&`, the ROS2Transform cosntructor still has to make a copy.
So total operations = 2 copies

Now if the ROS2Transform constructor had the signature of `AZStd::string`, the `AZStd::move(transientParentFrame)` and `AZStd::move(transientChildFrame)` expressions would cause the move constructor of `AZStd::string` to be invoked when calling the constructor.
Afterwards inside of the constructor, the `parentFrame` and `childFrame` parameters can then be moved into the member variables. In this case there are 4 move operations
So total operation = 4 moves

In this scenario `AZStd::string` signature wins.
4 moves are better than 2 copies, since a move only has to copy the data pointers, size and capacity values, while a copy has to allocate a block of memory, copy over the string data + copy over the size and capacity values

Code/Source/Frame/NamespaceConfiguration.cpp
lumberyard-employee-dm: Actually that would would result in a deallocation of `m_entityName`.
Once the move-assignment operation occurs, the `entityName` string data, size and capacity values "stolen" by the m_entityName members. At that point `entityName` is pointing to an empty AZStd::string instance with no-allocated memory(The string data pointer is set to nullptr, the size and capacity values are 0).

Code/Source/Frame/NamespaceConfiguration.cpp
spham-amzn: Thanks. IMHO, the switches on enumerations makes it easier to read the logic flow in these situations.

Code/Source/Frame/NamespaceConfiguration.cpp
adamdbrw: I would then rather stay with unchanged version here.

Code/Source/Frame/ROS2Transform.cpp
adamdbrw: I can change this constructor

spham-amzn (approved):

Code/Source/Frame/NamespaceConfiguration.cpp
lumberyard-employee-dm: Sorry I mistyped in the first sentence.
Accepting an `AZStd::string` parameter by value and moving `entityName` parameter into `m_entityName` would result in 0 de-allocations and a single move.

The `const AZStd::string&` version is guaranteed to have at least one copy and a potential second copy and de-allocation if a const char* is passed to `PopulateNamespace`
i.e
```
NamespaceConfiguration config;
config.PopulateNamespace("foo"); 
```
That results in copy "foo" to an AZStd::string variable in the call and then inside of the function during the assignment.
When the function completes, the "copy of "foo" also has to be destructed.
So in that case there are 2 copies and 1 destruction

lumberyard-employee-dm (approved):

Code/Source/Frame/NamespaceConfiguration.cpp
adamdbrw: What I meant is a situation when **m_entityName** already has a value (its own storage allocated). I am not familiar with AZStd::string but with std::string I believe that would be the case. I did not question what happens to entityName, only what happens to m_entityName.

adamdbrw: In any case, I will push a commit tomorrow, gotta rest a bit now :]

lumberyard-employee-dm: Yes, the `m_entityName` would result in a deallocation on move assignment operator in the [AZStd::string class](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/string/string.h#L295-L309) if it contained a previous value.
That being said it is balanced against the fact that it wouldn't need to allocate, which is a potential possibility of the copied string is larger than the existing capacity of `m_entityName`

adamdbrw: Additional consideration is that frame names are usually quite short and SSO is a thing: https://stackoverflow.com/questions/21694302/what-are-the-mechanics-of-short-string-optimization-in-libc

lumberyard-employee-dm: Ok, in that case the AZStd::string implementation, use the short string optimization for strings that are [22 characters](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/string/string.h#L311-L327) or less.
In that case it copies over the string data

PR #31, opened 2022-04-29
FBX files generator with basic functionalities

mdrwiega:
Currently, the generator is able only to generate cube objects with specific material and save it to file as ASCII FBX file.
It has been checked with Asset Processor that the generated file is correct and could be loaded to O3DE simulation.


Code/Source/URDF/FbxGenerator.cpp
```diff
@@ -0,0 +1,425 @@
...
+            properties.AddChild(Node("P", {"AmbientColor", "ColorRGB", "Color", "", 0, 0, 0}));
+            properties.AddChild(Node("P", {"DefaultCamera", "KString", "", "", "Producer Perspective"}));
+            properties.AddChild(Node("P", {"TimeMode", "enum", "", "", 11}));
+            properties.AddChild(Node("P", {"TimeProtocol", "enum", "", "", 2}));
+            properties.AddChild(Node("P", {"SnapOnFrameMode", "enum", "", "", 0}));
+            properties.AddChild(Node("P", {"TimeSpanStart", "KTime", "Time", "", 1924423250}));
+            properties.AddChild(Node("P", {"TimeSpanStop", "KTime", "Time", "", 1924423250}));
+            properties.AddChild(Node("P", {"CustomFrameRate", "double", "Number", "", -1}));
+            properties.AddChild(Node("P", {"TimeMarker", "Compound", "", ""}));
+            properties.AddChild(Node("P", {"CurrentTimeMarker", "int", "Integer", "", -1}));
```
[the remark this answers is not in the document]
mdrwiega: IMHO moving this to a config file will complicate the implementation (because then some kind of nodes reading method is necessary, for example, FBX reader). Also, this is rather a constant part that shouldn't be changed by us.

But if you think it's necessary, then I would propose creating a default/config FBX file only with this header which will be loaded and used in generated files. I thought about JSON but then the additional mapping between FBX constant parameters and JSON parameters is necessary (I'm afraid it could be confusing because of the number of parameters). Anyway, I think the FBX reader is not trivial (even if the structure is created), so it could take some time.

So, what do you think?

Code/Source/URDF/FbxGenerator.cpp
```diff
@@ -0,0 +1,425 @@
...
+            model.AddChild("Count", 1);
+            definitions.AddChild(std::move(model));
+
+            Node geometry("ObjectType", {"Geometry"});
+            geometry.AddChild("Count", 1);
+            definitions.AddChild(std::move(geometry));
+
+            Node material("ObjectType", {"Material"});
+            material.AddChild("Count", 1);
+            definitions.AddChild(std::move(material));
```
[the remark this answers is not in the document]
mdrwiega: Ok, I will add a method for example `AddNodeWithOneChild` or something like that. I haven't done it before because I wasn't sure if it's not more readable this way.

Code/Source/URDF/FbxGenerator.cpp
```diff
@@ -0,0 +1,425 @@
...
+            layerElementUV.AddChild(uv);
+
+            Node uvIndex("UVIndex", {RawString("*28")});
+            uvIndex.AddChild(Node("a", {0,1,3,2,2,3,5,4,4,5,7,6,6,7,9,8,1,10,11,3,12,0,2,13}));
+            layerElementUV.AddChild(uvIndex);
+
+            geometry.AddChild(layerElementUV);
+
+            return NodeWithId(geometryId, geometry);
+        }
```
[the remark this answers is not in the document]
mdrwiega: Yes, agree. It's even worse, many of them are not documented (FBX documentation is poor - it's intended to use with Autodesk library).

Code/Source/URDF/FbxGenerator.cpp
```diff
@@ -0,0 +1,425 @@
...
+            properties.AddChild(
+                Node("P", {"LastSaved|ApplicationName", "KString", "", "", "O3DE"}));
+            properties.AddChild(
+                Node("P", {"LastSaved|ApplicationVersion", "KString", "", "", "2022"}));
+            properties.AddChild(
+                Node("P", {"LastSaved|DateTime_GMT", "DateTime", "", "", "01/01/2022 00:00:00.000"}));
+            properties.AddChild(
+                Node("P", {"Original|ApplicationActiveProject", "KString", "", "", "/dummy_path.fbx"}));
+            properties.AddChild(
+                Node("P", {"Original|ApplicationNativeFile", "KString", "", "", "/dummy_path.fbx"}));
```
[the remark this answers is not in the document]
mdrwiega: I have answered above (on similar suggestion).

Code/Source/URDF/FbxNode.cpp
```diff
@@ -0,0 +1,161 @@
...
+                ss << " {\n";
+
+                if(m_children.size() > 0)
+                {
+                    for(auto node : m_children)
+                    {
+                        ss << node.ToString(nodeDepth + 1);
+                    }
+                }
+                ss << offset + "}\n";
```
spham-amzn, line 150:
```suggestion
                ss << offset << "}\n";
```

Code/Source/URDF/FbxNode.cpp
```diff
@@ -0,0 +1,161 @@
...
+            for (int i = 0; i < nodeDepth; ++i)
+                offset += "  ";
+
+            // Write name
+            ss << offset << m_name << ": ";
+
+            if (!HasProperties() && !HasChildren())
+            {
+                ss << " {\n";
+                ss << offset + "}";
```
spham-amzn, line 119:
```suggestion
                ss << offset << "}";
```

Code/Source/URDF/FbxNode.cpp
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        std::string Node::ToString(int nodeDepth) const
+        {
+            std::stringstream ss;
+            std::string offset;
+
+            // Calculate offset
+            for (int i = 0; i < nodeDepth; ++i)
+                offset += "  ";
```
spham-amzn, line 111: FYI:
You can use the std::setfill and std::setw into a another string stream
(You will need to add ```#include <iomanip>``` at the top of this file)
```suggestion
            // Calculate offset
            std::stringstream offset_stream;
            offset_stream << std::setfill('  ') << std::setw(nodeDepth) << "";
            std::string offset = offset_stream.str();
```

Code/Source/URDF/FbxNode.cpp
```diff
@@ -0,0 +1,161 @@
...
+            }
+            else if (a.type() == typeid(std::string))
+            {
+                return std::string("\"") + std::any_cast<std::string>(a) + std::string("\"");
+            }
+            else if (a.type() == typeid(RawString))
+            {
+                return std::any_cast<RawString>(a).data;
+            }
+
```
spham-amzn, line 50: Should put an assert here or something to catch any type that is not being handled, or at least a warning

Code/Tests/UrdfToFbxConverterTest.cpp
```diff
@@ -0,0 +1,55 @@
...
+                "      </geometry>"
+                "      <material name=\"black\"/>"
+                "    </visual>"
+                "    <collision>"
+                "      <geometry>"
+                "        <box size=\"1.0 2.0 1.0\"/>"
+                "      </geometry>"
+                "    </collision>"
+                "  </link>"
+                "</robot>";
```
spham-amzn, line 39: Without any linefeeds (\n) in the strings here, there is no point indenting / pretty printing the xml here. If the intent is to make it readable, I suggest adding '\n' to the end of each line

mdrwiega: Thank you a lot for your suggestion. This XML is intended only for the testing of the URDF to FBX conversion and it won't be saved or printed anywhere, so IMHO the pretty-printing is not necessary. Only the output from the converter will be analyzed (which is the FBX string).

Code/Source/URDF/UrdfToFbxConverter.h
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <string>
```
spham-amzn, line 11: It would be better to use AZStd::string instead of std::string since the rest of the engine and components depend on it. (and AZStd:: in general)

Code/Source/URDF/Constants.h
```diff
@@ -0,0 +1,72 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <string>
```
spham-amzn, line 11: It would be better to use AZStd::string instead of std::string since the rest of the engine and components depend on it. (and AZStd:: in general)

Code/Source/URDF/Constants.h
```diff
@@ -0,0 +1,72 @@
...
+    //! TODO: Could be moved to configuration file
+    namespace Fbx
+    {
+        struct Constants
+        {
+            struct FbxHeader
+            {
+                inline static const int headerVersion = 1003;
+                inline static const int fileVersion = 7500;
+                inline static const std::string creatorName = "O3DE ROS2 Gem";
```
spham-amzn, line 26: If you make the change to use AZStd::string instead of std::string, then you should not use AZStd::string as static variables due to timing of its allocator availability. Instead, you should use string constants instead:
```suggestion
                constexpr char* creatorName = "O3DE ROS2 Gem";
```

Code/Source/URDF/Constants.h
```diff
@@ -0,0 +1,72 @@
...
+    ////////////////////////////////////////////////////////////////////////////////////////////////////
+    //! The struct contains all constant parameters used in FBX files.
+    //! TODO: Could be moved to configuration file
+    namespace Fbx
+    {
+        struct Constants
+        {
+            struct FbxHeader
+            {
+                inline static const int headerVersion = 1003;
```
spham-amzn, line 24: ```inline``` is not necessary, and you can use constexpr instead for all the constants
```suggestion
                static constexpr int headerVersion = 1003;
```

mdrwiega: IMHO not a big difference but yes, probably `constexpr` a bit better expresses the intention of compile-time constant usage.

Code/Source/URDF/Constants.h
mdrwiega: Thanks for the suggestion. I'll consider the change to AZStd;

Code/Source/URDF/Constants.h
mdrwiega: As above. I'll consider the change to AZStd::string.

spham-amzn (approved):

spham-amzn (approved):

Code/Source/URDF/FbxNode.cpp
```diff
@@ -12,37 +12,38 @@
...
-#include <AzCore/Debug/Trace.h>
+#include <AzCore/std/string/string.h>
+#include <AzCore/Console/Console.h>
 
 namespace ROS2
 {
     namespace Fbx
     {
-        std::string AnyToString(const std::any & a)
+        AZStd::string AnyToString(const std::any & a)
```
spham-amzn, line 22: I'm so sorry, I meant to delete this suggestion because I realized AZStd::string does not implement to_string, so no need to do the unnecessary conversion

PR #35, opened 2022-05-04
Topic validation

### LINKED ISSUE #7: Validate ros2 names
Only a subset of possible strings is valid for ros2 names for topics, frames and namespaces. Currently, when invalid names are inputted, we get no warning, but a runtime exception occurs when creating ros2 entities (e.g. publisher with invalid topic). We would like to improve this by lifting ros2 validation up to when the user inputs names.

We want to validate following values:
- [x] Namespaces
- [x] Topic names
- [x] Frame names

adamdbrw:
This fixes #7.

Important: make sure your o3de version includes recent fixes to ChangeValidate: 
https://github.com/o3de/o3de/pull/9409
Otherwise this will crash.


adamdbrw:
This fixes #7.

Important: make sure your o3de version includes recent fixes to ChangeValidate: 
https://github.com/o3de/o3de/pull/9409
Otherwise this will crash.

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,87 @@ namespace ROS2
...
-        return AZStd::regex_replace(input, ros2Disallowedlist, "_");
+        rosified = AZStd::regex_replace(rosified, ros2Disallowedlist, stringToReplaceViolations);
+
+        if (std::isdigit(rosified[0]) || (input[0] != underscore && rosified[0] == underscore))
+        {   // Prepend "o3de_" if it would otherwise start with a number (which would violate ros2 name requirements)
+            // Also, starting with '_' is not desired unless explicit. Topics/namespaces/parameters starting with "_" are hidden by default.
+            const AZStd::string prependToNumberStart = "o3de_";
+            rosified = prependToNumberStart + rosified;
+        }
+        return rosified;
```
pijaro, line 56: I think we should throw some feedback to the user if the `RosifyName` substantially changes the topic. Maybe some log warn?

PR #39, opened 2022-05-05
Dynamic transform fix

### LINKED ISSUE #37: Dynamic transforms are not published correctly
Transforms cause conflict on message types with RViz2. TransformBroadcaster should be used instead of a publisher. 
Also, child frame is empty (this was caused by an untested change to std::move and C&P).

adamdbrw:
Fixes #37

Transforms now work correctly with rviz2. 

### DIFF
--- a/Code/Source/Frame/ROS2Transform.cpp
+++ b/Code/Source/Frame/ROS2Transform.cpp
@@ -32,12 +32,11 @@ namespace ROS2
 
     void ROS2Transform::Publish(const AZ::Transform& transform)
     {
-        static bool isPublished = false;
-        if (isPublished && !m_isDynamic)
+        if (m_isPublished && !m_isDynamic)
         {   // Only publish static transforms once
             return;
         }
         ROS2Interface::Get()->BroadcastTransform(CreateTransformMessage(transform), m_isDynamic);
-        isPublished = true;
+        m_isPublished = true;
     }
 }  // namespace ROS2
--- a/Code/Source/Frame/ROS2Transform.cpp
+++ b/Code/Source/Frame/ROS2Transform.cpp
@@ -14,59 +14,9 @@
 
 namespace ROS2
 {
-    namespace Internal
-    {
-        class StaticTransformPublisher : public TransformPublisher
-        {
-        public:
-            void Publish(const geometry_msgs::msg::TransformStamped& transformMessage) override
-            {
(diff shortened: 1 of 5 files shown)


Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -81,6 +32,12 @@ namespace ROS2
 
     void ROS2Transform::Publish(const AZ::Transform& transform)
     {
-        m_transformPublisher->Publish(CreateTransformMessage(transform));
+        static bool isPublished = false;
```
spham-amzn, line 35: I think ```isPublished``` should be moved to a private member of ROS2Transfrom class instead of declaring it static in this function

lumberyard-employee-dm: Especially if there can be multiple ROS2Transform instances.

Code/Source/ROS2SystemComponent.h
```diff
@@ -65,6 +66,7 @@ namespace ROS2
     private:
         std::shared_ptr<rclcpp::Node> m_ros2Node;
         AZStd::shared_ptr<rclcpp::executors::SingleThreadedExecutor> m_executor;
+        AZStd::unique_ptr<tf2_ros::TransformBroadcaster> m_dynamicTFBroadcaster;
         AZStd::unique_ptr<tf2_ros::StaticTransformBroadcaster> m_staticTFBroadcaster;
```
lumberyard-employee-dm, line 70: You could technically make an [AZStd::variant](https://en.cppreference.com/w/cpp/utility/variant) that stores both a TransformBroadcaster and a StaticTransformBroadcaster to replacate a type safe union.
i.e
```
using TransformBroadcaster = AZStd::varaiant<AZStd::unique_ptr<tf2_ros::StaticTransformBroadcaster>, AZStd::unique_ptr<tf2_ros::TransformBroadcaster>>;
TransformBroadcaster m_staticTFBroadcaster;
```

adamdbrw: Both static and dynamic broadcaster are always needed (each robot will typically have both static and dynamic frame transforms).

lumberyard-employee-dm: Ok thanks for the info.

lumberyard-employee-dm (approved):

PR #42, opened 2022-05-06
Fixed lidar orientation and points frame

pijaro:
Fixed:
- lidar beam directions
- results in local frame

### DIFF
--- a/Code/Source/Lidar/LidarTemplateUtils.cpp
+++ b/Code/Source/Lidar/LidarTemplateUtils.cpp
@@ -53,7 +53,8 @@ namespace ROS2
     }
 
     // TODO - lidars in reality do not have uniform distributions - populating needs to be defined per model
-    AZStd::vector<AZ::Vector3> LidarTemplateUtils::PopulateRayDirections(LidarTemplate::LidarModel model)
+    AZStd::vector<AZ::Vector3> LidarTemplateUtils::PopulateRayDirections(LidarTemplate::LidarModel model,
+                                                                         const AZ::Vector3& rootRotation)
     {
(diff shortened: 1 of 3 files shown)


Code/Source/Lidar/LidarTemplateUtils.h
```diff
@@ -17,6 +17,7 @@ namespace ROS2
     public:
         static LidarTemplate GetTemplate(LidarTemplate::LidarModel model);
         static size_t TotalPointCount(const LidarTemplate& t);
-        static AZStd::vector<AZ::Vector3> PopulateRayDirections(LidarTemplate::LidarModel model);
+        static AZStd::vector<AZ::Vector3> PopulateRayDirections(LidarTemplate::LidarModel model,
```
adamdbrw, line 20: could use a comment that angles are euler, radians

Code/Source/Lidar/LidarTemplateUtils.cpp
```diff
@@ -74,8 +75,8 @@ namespace ROS2
             for (int layer = 0; layer < lidarTemplate.m_layers; layer++)
             {
                 // roll is equal to 0, so it's skipped in the calculations
```
adamdbrw, line 77: we should not leave out roll transformation

pijaro: Agreed, but not the scope of this PR. Most probably we will move towards quaternions anyway.

Code/Source/Lidar/LidarTemplateUtils.h
pijaro: added comment

PR #53, opened 2022-05-12
Topic validation (#35-main)

adamdbrw:
Validation of ROS2 names:
- validate topics and namespaces
- rosify name now deals with names starting with numbers
- rosify name now deals with names starting with underscore. Warns if explicit, prepends o3de_ otherwise

### DIFF
--- a/Code/Source/Utilities/ROS2Names.cpp
+++ b/Code/Source/Utilities/ROS2Names.cpp
@@ -86,7 +86,7 @@ namespace ROS2
         return AZ::Success();
     }
 
-    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateNamespaceField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateNamespaceField(void* newValue, const AZ::Uuid& valueType)
     {
         if (azrtti_typeid<AZStd::string>() != valueType)
         {
@@ -115,7 +115,7 @@ namespace ROS2
         return AZ::Success();
     }
 
-    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateTopicField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateTopicField(void* newValue, const AZ::Uuid& valueType)
     {
         if (azrtti_typeid<AZStd::string>() != valueType)
         {
--- a/Code/Source/Utilities/ROS2Names.cpp
+++ b/Code/Source/Utilities/ROS2Names.cpp
@@ -93,7 +93,7 @@ namespace ROS2
             return AZ::Failure(AZStd::string("Unexpected field type: the only valid input is a character string"));
         }
 
-        AZStd::string ros2Namespace(static_cast<const char*>(newValue));
+        const AZStd::string& ros2Namespace(*reinterpret_cast<const AZStd::string*>(newValue));
         return ValidateNamespace(ros2Namespace);
     }
 
@@ -122,7 +122,7 @@ namespace ROS2
             return AZ::Failure(AZStd::string("Unexpected field type: the only valid input is a character string"));
         }
 
-        AZStd::string topic(static_cast<const char*>(newValue));
+        const AZStd::string& topic(*reinterpret_cast<const AZStd::string*>(newValue));
         return ValidateTopic(topic);
     }
 }  // namespace ROS2
--- a/Code/Source/Utilities/ROS2Names.cpp
+++ b/Code/Source/Utilities/ROS2Names.cpp
@@ -8,6 +8,8 @@
 
 #include "ROS2Names.h"
 #include <AzCore/std/string/regex.h>
+#include <rcl/validate_topic_name.h>
+#include <rmw/validate_namespace.h>
 
 namespace ROS2
 {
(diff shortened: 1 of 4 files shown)


Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,92 @@ namespace ROS2
...
+    }
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateNamespace(const AZStd::string& ros2Namespace)
+    {
+        auto ros2GlobalizedNamespace = ros2Namespace;
+        const char namespacePrefix = '/';
+        if (ros2Namespace.find(namespacePrefix, 0) != 0)
+        {  // Prepend "/" if not included, this is done automatically by rclcpp so "/"-less namespaces are ok.
+            ros2GlobalizedNamespace = namespacePrefix + ros2Namespace;
+        }
```
lumberyard-employee-dm, line 71: You can use the [AZStd::string::starts_with](https://en.cppreference.com/w/cpp/string/basic_string/starts_with) function to determine if the string starts with a forward slash.

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,92 @@ namespace ROS2
...
+            return AZ::Failure(AZStd::string(rcl_topic_name_validation_result_string(validationResult)));
+        }
+
+        return AZ::Success();
+    }
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateTopicField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
+    {
+        AZStd::string topic(static_cast<const char*>(newValue));
+        return ValidateTopic(topic);
```
lumberyard-employee-dm, line 116: Seems like this should validate that the TypeId is `azrtti_typeid<const char>`

```
valueType == azrtti_typeid<const char>()
```

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,92 @@ namespace ROS2
...
+        }
+
+        return AZ::Success();
+    }
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateNamespaceField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
+    {
+        AZStd::string ros2Namespace(static_cast<const char*>(newValue));
+        return ValidateNamespace(ros2Namespace);
+    }
```
lumberyard-employee-dm, line 93: Repeat: we should validate the type to make sure it is a pointer to `const char`
```
valueType == azrtti_typeid<const char>()
```

Code/Source/Utilities/ROS2Names.cpp
adamdbrw: I am adding a check, but it seems that the actual valueType for this field is AZStd::string. Is this intended? 

Here are the values when called in the good scenario (valueType as called, AZStd::string type UUID, const char type UUID)
{03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}, {03AAAB3F-5C47-5A66-9EBC-D5FA4DB353C9}, {3AB0037F-AF8D-48CE-BCA0-A170D18B2C03}

I will commit changes to reflect this, but let me know if this is correct or going to change.

adamdbrw:
Review comments were addressed. @lumberyard-employee-dm can you take another look?

lumberyard-employee-dm:
> Review comments were addressed. @lumberyard-employee-dm can you take another look?

Yes, I am looking now

Code/Source/Utilities/ROS2Names.cpp
lumberyard-employee-dm: Ok if the Uuid is AZStd::string, that means that the void pointer is a pointer to an `AZStd::string`, not a `const char*`, so you will need to update the void* cast here.

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,102 @@ namespace ROS2
...
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateNamespaceField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
+    {
+        if (azrtti_typeid<AZStd::string>() != valueType)
+        {
+            return AZ::Failure(AZStd::string("Unexpected field type: the only valid input is a character string"));
+        }
+
+        AZStd::string ros2Namespace(static_cast<const char*>(newValue));
+        return ValidateNamespace(ros2Namespace);
```
lumberyard-employee-dm, line 97: Since the type id here is `AZStd::string`, the void pointer should be cast to an `AZStd::string*`

```suggestion
        return ValidateNamespace(*reinterpret_cast<AZStd::string*>(newValue));
```

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,102 @@ namespace ROS2
...
+    }
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateTopicField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
+    {
+        if (azrtti_typeid<AZStd::string>() != valueType)
+        {
+            return AZ::Failure(AZStd::string("Unexpected field type: the only valid input is a character string"));
+        }
+
+        AZStd::string topic(static_cast<const char*>(newValue));
```
lumberyard-employee-dm, line 125: repeat: If the `valueType` value matches `azrtti_typeid<AZStd::string>()`, then the void pointer type is `AZStd::string*`

```suggestion
        const AZStd::string& topic(*reinterpret_cast<AZStd::string*>(newValue));
```

adamdbrw:
Applied

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,102 @@ namespace ROS2
...
+
+        if (validationResult != RMW_NAMESPACE_VALID)
+        {
+            return AZ::Failure(AZStd::string(rmw_namespace_validation_result_string(validationResult)));
+        }
+
+        return AZ::Success();
+    }
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateNamespaceField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
```
spham-amzn, line 89: ```[[maybe_unused]]``` doesn't seem to be necessary here

Code/Source/Utilities/ROS2Names.cpp
```diff
@@ -25,8 +27,102 @@ namespace ROS2
...
+
+        if (RCL_TOPIC_NAME_VALID != validationResult)
+        {
+            return AZ::Failure(AZStd::string(rcl_topic_name_validation_result_string(validationResult)));
+        }
+
+        return AZ::Success();
+    }
+
+    AZ::Outcome<void, AZStd::string> ROS2Names::ValidateTopicField(void* newValue, [[maybe_unused]] const AZ::Uuid& valueType)
```
spham-amzn, line 118: ```[[maybe_unused]]``` doesn't seem to be necessary here

spham-amzn (approved):

lumberyard-employee-dm (approved):

PR #55, opened 2022-05-12
Lidar Component

adamdbrw:
- Lidar templates to pick from. Comes with just one for now: Generic Lidar.
- Raycaster based on PhysX scene queries
- Visualising data with AuxGeom
- Transformation from global to local frame
- Filling and publishing ros2 PointCloud2 message

Note: this is a first iteration of the lidar component.

### DIFF
--- a/Code/Source/Lidar/LidarRaycaster.cpp
+++ b/Code/Source/Lidar/LidarRaycaster.cpp
@@ -19,6 +19,7 @@ namespace ROS2
     {
         AZStd::vector<AZ::Vector3> results;
         AzPhysics::SceneQueryRequests requests;
+        requests.reserve(directions.size());
         for (const AZ::Vector3& direction : directions)
         {   // NOTE - performance-wise, consider reusing requests
             AZStd::shared_ptr<AzPhysics::RayCastRequest> request = AZStd::make_shared<AzPhysics::RayCastRequest>();
--- a/Code/Source/Lidar/LidarRaycaster.cpp
+++ b/Code/Source/Lidar/LidarRaycaster.cpp
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#include "LidarRaycaster.h"
+#include <AzFramework/Physics/PhysicsScene.h>
+#include <AzCore/Interface/Interface.h>
+#include <AzFramework/Physics/Common/PhysicsSceneQueries.h>
+#include <AzCore/std/smart_ptr/shared_ptr.h>
+#include <AzCore/std/smart_ptr/make_shared.h>
+
+namespace ROS2
+{
+    // A simplified, non-optimized first version. TODO - generalize results (fields)
+    AZStd::vector<AZ::Vector3> LidarRaycaster::PerformRaycast(const AZ::Vector3& start, const AZStd::vector<AZ::Vector3>& directions, float distance)
+    {
+        AZStd::vector<AZ::Vector3> results;
+        AzPhysics::SceneQueryRequests requests;
+        for (const AZ::Vector3& direction : directions)
+        {   // NOTE - performance-wise, consider reusing requests
+            AZStd::shared_ptr<AzPhysics::RayCastRequest> request = AZStd::make_shared<AzPhysics::RayCastRequest>();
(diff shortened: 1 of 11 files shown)


adamdbrw:
Note: to run with this component, first merge #53 and #54

Code/Source/Lidar/LidarRaycaster.cpp
```diff
@@ -0,0 +1,49 @@
...
+        AzPhysics::SceneQueryRequests requests;
+        for (const AZ::Vector3& direction : directions)
+        {   // NOTE - performance-wise, consider reusing requests
+            AZStd::shared_ptr<AzPhysics::RayCastRequest> request = AZStd::make_shared<AzPhysics::RayCastRequest>();
+            request->m_start = start;
+            request->m_direction = direction;
+            request->m_distance = distance;
+            request->m_reportMultipleHits = false;
+            requests.emplace_back(AZStd::move(request));
+        }
```
lumberyard-employee-dm, line 30: This could `reserve` the size of the `directions` vector for the `requests` vector to reduce the number of allocations.

Code/Source/Lidar/LidarTemplateUtils.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+    LidarTemplate LidarTemplateUtils::GetTemplate(LidarTemplate::LidarModel model)
+    {
+        static std::unordered_map<LidarTemplate::LidarModel, LidarTemplate> templates;
+
+        if (templates.empty())
+        {
+            LidarTemplate generic3DLidar =
+            {
+                .m_model = LidarTemplate::Generic3DLidar,
```
lumberyard-employee-dm, line 28: [Designated Initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers) isn't supported until C++20, so this code shouldn't be compiling.

Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -0,0 +1,159 @@
...
+        auto lidarTM = entityTransform->GetWorldTM();
+        lidarTM.Invert();
+
+        // TODO - improve performance
+        for (auto& point : m_lastScanResults)
+        {
+            point = lidarTM.TransformPoint(point);
+        }
+
+        memcpy(message.data.data(), m_lastScanResults.data(), message.data.size());
```
lumberyard-employee-dm, line 156: Is this missing a multiplication against `sizeof <SomeType>`?
It looks like the `message.data` type is some kind of vector, but is it a vector of a 1 byte type.

Code/Source/Lidar/LidarTemplateUtils.cpp
adamdbrw: I am using clang 12, it does support designated initializers (since clang 10).
https://en.cppreference.com/w/cpp/compiler_support.
However, I understand that this is not universal among supported platforms and compilers.
Default member initialization is inferior as it is relatively easy to confuse the order or miss an error when extending struct. But if it is all we have - I will switch to it.

Code/Source/Lidar/ROS2LidarSensorComponent.cpp
adamdbrw: It is correct as it is. The type for PointCloud2 message data field is specified as single-byte type (uint8): 
http://docs.ros.org/en/lunar/api/sensor_msgs/html/msg/PointCloud2.html

Code/Source/Lidar/LidarTemplateUtils.cpp
lumberyard-employee-dm: Default Member initialization is initializing the class members as part of the struct definition.
That is different from the member initializer list which is used as part of the constructor.

By [default member initialization](https://en.cppreference.com/w/cpp/language/data_members#Member_initialization) it would look like the following
```
struct LidarTemplate
{
       LidarModel m_model;
        AZStd::string m_name;
        float m_minHAngle = 0.0f;
        float m_maxHAngle = 0.0f;
        float m_minVAngle = 0.0f;
        float m_maxVAngle = 0.0f;
        int m_layers = 0;
        int m_numberOfIncrements = 0;
        float m_maxRange = 0.0f;
}
```

adamdbrw: Oh I see, I thought you meant something else. This won't do as it does not help if there is more than one type of lidar.

adamdbrw:
@lumberyard-employee-dm I applied all the comments

lumberyard-employee-dm (approved):

PR #58, opened 2022-05-17
Robot control component

adamdbrw:
A robot control component, first approach:
- Can work with scripts or with a default implementation
- Comes with a default rigid body Twist control
- Comes with an example lua script
- Enables setting of topic and QoS


Code/Source/ROS2ModuleInterface.h
```diff
@@ -8,6 +8,7 @@
 #include <AzCore/Memory/SystemAllocator.h>
 #include <AzCore/Module/Module.h>
 #include <Frame/ROS2FrameComponent.h>
+#include <RobotControl/ROS2RobotControlComponent.h">
```
lumberyard-employee-dm, line 11: There is an additional `"` character in the include path that shouldn't be here

Code/Source/RobotControl/ControlConfiguration.cpp
```diff
@@ -0,0 +1,69 @@
+/*
+* Copyright (c) Contributors to the Open 3D Engine Project.
+* For complete copyright and license terms please see the LICENSE at the root of this distribution.
+*
+* SPDX-License-Identifier: Apache-2.0 OR MIT
+*
+*/
+
+#pragma once
```
lumberyard-employee-dm, line 9: The `#pragma once` in a cpp file will cause a compiler warning which will turn into a compile error due to `-Werror`

Code/Source/RobotControl/ROS2RobotControlComponent.cpp
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
```
lumberyard-employee-dm, line 8: Repeat: `#pragma once` needs to be removed

Code/Source/RobotControl/RobotConfiguration.cpp
```diff
@@ -0,0 +1,93 @@
+/*
+* Copyright (c) Contributors to the Open 3D Engine Project.
+* For complete copyright and license terms please see the LICENSE at the root of this distribution.
+*
+* SPDX-License-Identifier: Apache-2.0 OR MIT
+*
+*/
+
+#pragma once
```
lumberyard-employee-dm, line 9: repeat: `#pragma once` needs to be removed in order to compile successfully

lumberyard-employee-dm (changes requested):

Code/Source/RobotControl/RobotControl.h
```diff
@@ -0,0 +1,86 @@
...
+                auto ros2Frame = entity->FindComponent<ROS2FrameComponent>();
+                AZStd::string namespacedTopic = ROS2Names::GetNamespacedName(
+                        ros2Frame->GetNamespace(),
+                        m_controlConfiguration.m_topic);
+
+                auto ros2Node = ROS2Interface::Get()->GetNode();
+                m_controlSubscription = ros2Node->create_subscription<T>(
+                    namespacedTopic.data(),
+                    m_controlConfiguration.m_qos.GetQoS(),
+                    std::bind(&RobotControl<T>::OnControlMessage, this, std::placeholders::_1));
```
lumberyard-employee-dm, line 51: FYI: Using a lambda results in more [efficient code](https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-bind.html) than using std::bind.

```suggestion
                   [this](const T& message) { OnControlMessage(message); });
```

Code/Source/ROS2ModuleInterface.h
adamdbrw: My bad - a merge conflict between <> and "" includes was resolved with this rather ugly child!

Code/Source/RobotControl/ControlConfiguration.cpp
adamdbrw: @pijaro

adamdbrw:
@pijaro for review changes for this PR, please apply them as a commit to develop first, check if everything works and then cherry-pick into this branch.

pijaro:
@lumberyard-employee-dm Thanks for the review :bow:, I applied all of the suggested changes.

lumberyard-employee-dm (approved):

PR #65, opened 2022-05-24
ROS2 camera sensor (#65)

adamdbrw:
Camera sensor implementation

Expected features:
- [x] image available through ROS2 topic
- [x] rendering to texture
- [x] rendering from specified transform
- [x] multi-camera support
- [ ] performance improvement (required feature merge to o3de develop)


Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -0,0 +1,53 @@
...
+#include "CameraSensor.h"
+
+namespace ROS2
+{
+    //! ROS2 Camera sensor component class
+    //! Allows to place camera sensor in the scene
+    //! Can be parametrized with following values:
+    //!   - camera name
+    //!   - camera image width and height in pixels
+    //!   - camera vertical field of view in degrees
```
prybicki: If my Unity-based intuition is correct, it would be better to write:
`Allows turning a game object [object on the scene) into a camera sensor` or something in similar meaning.
In essence, the current phrasing may tell someone inexperienced that they need to drag and drop a camera sensor component into the scene (which is true for prefabs, not code components), where in fact you need to add the component to an existing object on the scene.

piotr-zyskowski-rai: Totally agree. One small note: game objects are called entities in o3de so the final sentence would be
```
Allows turning an entity into a camera sensor
```

Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -0,0 +1,53 @@
...
+        int m_width = 640;
+        int m_height = 480;
+        AZStd::string m_cameraName = "camera";
+
+        void FrequencyTick() override;
+
+        std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::Image>> m_imagePublisher;
+        CameraSensor m_cameraSensor;
+
+        AZStd::string m_frameName;
```
prybicki: How about some ROS2 QoS settings?

Code/Source/Camera/ROS2CameraSensorComponent.cpp
```diff
@@ -0,0 +1,99 @@
...
+        m_cameraSensor.RequestFrame(transform, [this](const AZ::RPI::AttachmentReadback::ReadbackResult& result) {
+            const AZ::RHI::ImageDescriptor& descriptor = result.m_imageDescriptor;
+
+            sensor_msgs::msg::Image message;
+            message.encoding = sensor_msgs::image_encodings::RGBA8;
+
+            message.width = descriptor.m_size.m_width;
+            message.height = descriptor.m_size.m_height;
+            message.data = std::vector<uint8_t>(result.m_dataBuffer->data(), result.m_dataBuffer->data() + result.m_dataBuffer->size());
+            message.header.frame_id = m_cameraName.c_str();
```
prybicki: Hmm, perhaps `m_frameName` was meant to be used here?

Code/Source/Camera/ROS2CameraSensorComponent.h
adamdbrw: QoS are handled through Sensor abstraction

Code/Source/Camera/ROS2CameraSensorComponent.cpp
adamdbrw: Yes, Frame id should come from the Frame Component

adamdbrw: Please make sure this component respects Sensor and Frame design as other components

Code/Source/Camera/ROS2CameraSensorComponent.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        auto* serialize = azrtti_cast<AZ::SerializeContext*>(context);
+        if (serialize)
+        {
+            serialize->Class<ROS2CameraSensorComponent, ROS2SensorComponent>()
+                    ->Version(1)
+                    ->Field("CameraName", &ROS2CameraSensorComponent::m_cameraName)
+                    ->Field("VerticalFieldOfViewDeg", &ROS2CameraSensorComponent::m_VerticalFieldOfViewDeg)
+                    ->Field("Width", &ROS2CameraSensorComponent::m_width)
+                    ->Field("Height", &ROS2CameraSensorComponent::m_height);
```
prybicki, line 47: `m_frameName`seems to be omitted in the serialization. Is this intended?

Code/Source/Camera/ROS2CameraSensorComponent.cpp
piotr-zyskowski-rai: `m_frameName` should not be present at all. It should be taken from the Frame component actually. I'll fix that

Code/Source/Camera/ROS2CameraSensorComponent.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+            AZ::EditContext* ec = serialize->GetEditContext();
+            if (ec) {
+                ec->Class<ROS2CameraSensorComponent>("ROS2 Camera Sensor", "[Simple Camera component]")
```
prybicki: As discussed, simple is very subjective ;)

Code/Source/Camera/ROS2CameraSensorComponent.cpp
piotr-zyskowski-rai: Intended as it should not exist at all :)

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+#include <chrono>
+
+class Entity;
+
+namespace ROS2 {
+
+    //! Structure containing all information required to create the camera sensor
+    struct CameraSensorDescription {
+        //! Constructor to create the description
+        //! @param cameraName - name of the camera; used to differentiate the cameras in multi-camera setup
```
prybicki:
```suggestion
        //! @param cameraName - name of the camera; used to differentiate cameras in a multi-camera setup
```

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+        //! @param verticalFov - vertical field of view of camera sensor
+        //! @param width - camera image width in pixels
+        //! @param height - camera image height in pixels
+        CameraSensorDescription(const AZStd::string& cameraName, float verticalFov, int width, int height);
+        CameraSensorDescription() = delete;
```
prybicki: Hmm, not sure but I guess it is implicitly deleted.

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+
+        const float verticalFieldOfViewDeg; //!< cameras vertical vield of view
```
prybicki:
```suggestion
        const float verticalFieldOfViewDeg; //!< camera vertical field of view
```

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+        const int width;                    //!< camera image width in pixels
+        const int height;                   //!< camera image height in pixels
+        const AZStd::string cameraName;     //!< camera nam eto differeniate the camera in mulit-camera setup
```
prybicki:
```suggestion
        const AZStd::string cameraName;     //!< camera name to differentiate cameras in a multi-camera setup
```

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+
+        const float aspectRatio;              //!< camera image aspect ration; derived from widht and height
```
prybicki:
```suggestion
        const float aspectRatio;              //!< camera image aspect ratio; equal to (width / height)
```

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+        const AZ::Matrix4x4 viewToClipMatrix; //!< camera view to clip space transform matrix; derived from other parameters
+
+    private:
+        AZ::Matrix4x4 MakeViewToClipMatrix() const;
+        void validateParameters() const;
+    };
+
+    //! Class to create camera sensor using Atom renderer
+    //! It creates dedicated rendering pipeline for each camera
```
prybicki, line 40: Are implications of many rendering pipelines known? E.g. what resources are used per RP? Could cameras share the rendering pipeline?

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+    class CameraSensor {
+    public:
+        //! Function to initialize rendering pipeline for the camera sensor
```
prybicki:
```suggestion
        //! Initializes rendering pipeline for the camera sensor
```
We all know these are functions :)

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+        //! @param cameraSensorDescription - camera sensor description used to create camera pipeline
+        void InitializePipeline(const CameraSensorDescription& cameraSensorDescription);
+
+        //! Function to deinitialize rendering pipeline for the camera sensor
+        void DestroyPipeline();
+
+        //! Function requiesting frame from rendering pipeline
```
prybicki:
```suggestion
        //! Function requesting frame from the rendering pipeline
```

Code/Source/Camera/CameraSensor.h
```diff
@@ -0,0 +1,67 @@
...
+        void RequestFrame(const AZ::Transform& cameraPose, std::function<void(const AZ::RPI::AttachmentReadback::ReadbackResult& result)> callback);
+
+    private:
+        AZStd::vector<AZStd::string> m_passHierarchy;
+        AZ::RPI::RenderPipelinePtr m_pipeline;
+        AZ::RPI::ViewPtr m_view;
+
+        AZ::RPI::Scene* m_scene = nullptr;
+
+        bool m_isInitialized = false;
```
prybicki: I may not have the full picture, however this is strongly against my intuition of good practices. In C++, the sole existence of an object should establish its invariants (RAII). Using std::optional<CameraSensor> in the caller code may help to achieve this.

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -0,0 +1,110 @@
...
+    , width(width)
+    , height(height)
+    , cameraName(cameraName)
+    , aspectRatio(static_cast<float>(width)/static_cast<float>(height))
+    , viewToClipMatrix(MakeViewToClipMatrix()) {
+        validateParameters();
+    }
+
+    AZ::Matrix4x4 CameraSensorDescription::MakeViewToClipMatrix() const {
+        const float nearDist = 0.1f, farDist = 100.0f;
```
prybicki, line 38: In the future, this might be an (advanced) configurable parameter of the component, I'd leave a TODO here (or add an internal task to track this).

Code/Source/Camera/CameraSensor.h
piotr-zyskowski-rai: It is Atom's internal thing. Not much knowledge here. But people from Atom's team agreed that a separate pipeline is a way to go when implementing a camera sensor.
The cameras almost certainly couldn't share the rendering pipeline fully as they are rendering from different poses. Some parts could maybe be shared but that's the tale for another time.

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -0,0 +1,110 @@
...
+        AZ::Matrix4x4 localViewToClipMatrix;
+        AZ::MakePerspectiveFovMatrixRH(localViewToClipMatrix, AZ::DegToRad(verticalFieldOfViewDeg), aspectRatio, nearDist, farDist, true);
+        return localViewToClipMatrix;
+    }
+
+    void CameraSensorDescription::validateParameters() const {
+        AZ_Assert(verticalFieldOfViewDeg > 0.0f && verticalFieldOfViewDeg < 180.0f, "Vertical fov should be in range 0.0 < FoV < 180.0 degrees");
```
prybicki, line 45: Given aspectRatio > 1.0, the code seem to be OK with horizontalFieldOfView being > 180.0f. Is that OK?

Code/Source/Camera/CameraSensor.cpp
piotr-zyskowski-rai: I treated it as an internal rendering parameter and not for the user to decide. As an advanced property maybe, yes but I do not see the point of marking it with TODO. If such a necessity arise, it will be implemented anyway

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -0,0 +1,110 @@
...
+        m_view->SetViewToClipMatrix(cameraSensorDescription.viewToClipMatrix);
+        m_scene = AZ::RPI::RPISystemInterface::Get()->GetSceneByName(AZ::Name("Main"));
+
+        AZStd::string pipelineName = cameraSensorDescription.cameraName + "Pipeline";
+
+        AZ::RPI::RenderPipelineDescriptor pipelineDesc;
+        pipelineDesc.m_mainViewTagName = "MainCamera";
+        pipelineDesc.m_name = pipelineName;
+        pipelineDesc.m_rootPassTemplate = "MainPipelineRenderToTexture";
+        pipelineDesc.m_renderSettings.m_multisampleState.m_samples = 4;
```
prybicki, line 64: This is likely to affect performance, IMHO it's worth considering exposing it to users.

Code/Source/Camera/CameraSensor.cpp
piotr-zyskowski-rai: I was there too. horizontalFoV is not aspectRatio * verticalFieldOfView

Code/Source/Camera/CameraCaptureScheduler.h
```diff
@@ -0,0 +1,68 @@
...
+#include <Clock/SimulationClock.h>
+
+#include <Atom/RPI.Public/RenderPipeline.h>
+
+#include <list>
+
+namespace ROS2 {
+
+//! Singleton class to initialize and synchronize the frame captures for multiple cameras
+//! It supports following features:
```
prybicki:
```suggestion
//! It supports the following features:
```

Code/Source/Camera/CameraCaptureScheduler.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <Clock/SimulationClock.h>
+
+#include <Atom/RPI.Public/RenderPipeline.h>
+
+#include <list>
+
+namespace ROS2 {
+
+//! Singleton class to initialize and synchronize the frame captures for multiple cameras
```
prybicki:
```suggestion
//! Singleton class to initialize and synchronize frame captures for multiple cameras
```

Code/Source/Camera/CameraSensor.cpp
piotr-zyskowski-rai: True. But I would rather consider it to be configurable via pipeline template definition, not from editor parameters. I'll look into it

Code/Source/Camera/CameraCaptureScheduler.h
```diff
@@ -0,0 +1,68 @@
...
+//! It supports following features:
+//! - only one frame is rendered and captured at a time with waiting for capture finished notification
```
prybicki: This sentence is hard to understand at first reading.

Code/Source/Camera/CameraCaptureScheduler.h
```diff
@@ -0,0 +1,68 @@
...
+    static CameraCaptureScheduler& Get();
+
+    //! Requests frame rendering if applicable, sets callback and adds to rendering queue is necessary
+    //! Calling this function does not guarantee start of the rendering.
+    //!
+    //! @note Pass hierarchy is expected to have two elements, where first is the unique pipeline name
+    bool TryRequestFrame(const AZStd::vector <AZStd::string>& passHierarchy,
+                         std::function<void(
+                              const AZ::RPI::AttachmentReadback::ReadbackResult &result)> frameRenderedCallback,
+                         AZ::RPI::RenderPipelinePtr pipeline);
```
prybicki, line 40:
```suggestion
    bool TryRequestFrame(const AZStd::vector<AZStd::string>& passHierarchy,
                         std::function<void(
                              const AZ::RPI::AttachmentReadback::ReadbackResult& result)> frameRenderedCallback,
                         AZ::RPI::RenderPipelinePtr pipeline);
```

Code/Source/Camera/CameraCaptureScheduler.h
```diff
@@ -0,0 +1,68 @@
...
+
+private:
+    void OnCaptureFinished(AZ::Render::FrameCaptureResult result, const AZStd::string &info) override;
```
prybicki, line 43:
```suggestion
    void OnCaptureFinished(AZ::Render::FrameCaptureResult result, const AZStd::string& info) override;
```

Code/Source/Camera/CameraCaptureScheduler.h
piotr-zyskowski-rai: split?
```suggestion
//! - only one frame is rendered and captured at a time 
//! - scheduler waits for a notification that capture is finished
```

Code/Source/Camera/CameraSensor.cpp
prybicki: Doesn't it mean that the camera won't show objects further than 100.0? It would be pretty sad if someone's use case is a large-scale Star Wars-like robot fleet.

Code/Source/Camera/CameraSensor.cpp
piotr-zyskowski-rai: Ok, this seems to be the place where this is configured. I'll leave this for now. It will take more hassle to do than I can give at the moment to do it properly. I'll fix that in the future (TODO imminent). This parameter should be exposed to the user in the same category as the near and far plane so I'll solve this later

PR #66, opened 2022-05-24
Documentation of ROS2Requests (ROS2Interface)

adamdbrw:

### DIFF
--- a/Code/Include/ROS2/ROS2Bus.h
+++ b/Code/Include/ROS2/ROS2Bus.h
@@ -43,6 +43,7 @@ namespace ROS2
         virtual builtin_interfaces::msg::Time GetROSTimestamp() const = 0;
 
         //! Send transformation between ROS2 frames.
+        //! @param t is a <a href="https://docs.ros2.org/latest/api/geometry_msgs/msg/TransformStamped.html">ROS2 TransformStamped message</a>.
         //! @param isDynamic controls whether a static or dynamic transform is sent. Static transforms are published
         //! only once and are to be used when the spatial relationship between two frames does not change.
         //! @note Transforms are already published by each ROS2FrameComponent.
--- a/Code/Include/ROS2/ROS2Bus.h
+++ b/Code/Include/ROS2/ROS2Bus.h
@@ -15,15 +15,38 @@
 
 namespace ROS2
 {
+    //! Interface to the central ROS2SystemComponent.
+    //! Use this API through ROS2Interface, for example:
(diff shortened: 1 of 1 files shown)


Code/Include/ROS2/ROS2Bus.h
```diff
@@ -15,15 +15,38 @@
...
+        //! @return Simulation time in ROS2 format. Time source is also valid with non-real time settings.
+        //! @note Make sure to set the use_sim_time parameter for ROS2 nodes which will use the simulation data.
         virtual builtin_interfaces::msg::Time GetROSTimestamp() const = 0;
+
+        //! Send transformation between ROS2 frames.
+        //! @param isDynamic controls whether a static or dynamic transform is sent. Static transforms are published
+        //! only once and are to be used when the spatial relationship between two frames does not change.
+        //! @note Transforms are already published by each ROS2FrameComponent.
+        //! Use this function directly only when default behavior of ROS2FrameComponent is not sufficient.
         virtual void BroadcastTransform(const geometry_msgs::msg::TransformStamped& t, bool isDynamic) const = 0;
```
lumberyard-employee-dm, line 50: Looks like the @param for the `t` parameter is missing.

lumberyard-employee-dm (commented):
Besides from the missing @param for the `t` parameter of BroadcastTransform, it the comment style fits with what is in the [API Ref Guidelines](https://github.com/o3de/sig-core/blob/main/governance/API-Ref-Guidelines-Update.md#how-should-i-format-my-code-comments)

Code/Include/ROS2/ROS2Bus.h
adamdbrw: that was on purpose. It has no additional context except what is already documented here: https://docs.ros2.org/latest/api/geometry_msgs/msg/TransformStamped.html. But this is also true for many other ros2 structs and types. I thought referring users to ros2 documentation as a part of top level docs would be better than providing refs for each of them in the comments, but I also think this could be up for a discussion.

lumberyard-employee-dm: Well from outside perspective, I think users would have hard time determining what the TransformStamped parameter should be set to.

lumberyard-employee-dm (approved):

piotr-zyskowski-rai (approved):
LGTM

adamdbrw:
Advances https://github.com/RobotecAI/o3de-ros2-gem/issues/70

PR #83, opened 2022-05-31
Add URDF to FBX conversion

mdrwiega:
Changes:
- handle materials
- created tests: with two links and joint and with simple robot
- depth first search in URDF links tree and add relations based on
joints
- switched to AZStd containers
- fixed memory allocation in tests

Signed-off-by: Michal Drwiega <michal.drwiega@robotec.ai>


piotr-zyskowski-rai (approved):
LGTM

PR #91, opened 2022-06-07
Lidar improvements (main)

adamdbrw:
- Generic Lidar now allows customization (as intended) of parameters in the Editor
- Lidars are now not bound to the default scene, instead, they take the owning Entity's scene.
- Lidars can now have a filtered (lidar-transparent) object. By default, it is their own entity. This prevent lidar from detecting only the inside of own box model.
- Some prefab improvements.
- Bugfix to frequency handling with OnTIck - replaced debug static with a member

### DIFF
--- a/Code/Source/Lidar/LidarTemplate.h
+++ b/Code/Source/Lidar/LidarTemplate.h
@@ -35,8 +35,8 @@ namespace ROS2
         float m_maxHAngle = 0.0f;
         float m_minVAngle = 0.0f;
         float m_maxVAngle = 0.0f;
-        int m_layers = 0;
-        int m_numberOfIncrements = 0;
+        unsigned int m_layers = 0;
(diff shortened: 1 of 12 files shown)


Code/Source/Lidar/LidarTemplate.h
```diff
@@ -35,8 +35,8 @@ namespace ROS2
         float m_maxHAngle = 0.0f;
         float m_minVAngle = 0.0f;
         float m_maxVAngle = 0.0f;
-        int m_layers = 0;
-        int m_numberOfIncrements = 0;
+        unsigned int m_layers = 0;
+        unsigned int m_numberOfIncrements = 0;
```
lumberyard-employee-dm, line 39: Suggestion: It would be probably be more future proof to use `uint32_t`

Issue #22, opened 2022-04-21
Global map handler


adamdbrw:
Global map handler would act as a singleton (another SystemComponent) and have several connected features and responsibilities:
1. Be configurable by user easily from the Editor (as for next points)
2. Hold map translation / rotation and placement in global frame (for example, GPS coordinates of origin)
3. Hold the name of global frame (e.g. "map", or "world"
4. To be discussed: hold static tf broadcaster (without knowledge of transforms, just exposing the publish interface). This is currently in ROS2SystemComponent.

adamdbrw:
Unnecesary for the demo - moving to General Backlog

PR #92, opened 2022-06-07
Changed root frame to odom (#86)

adamdbrw:
Change the global frame to "odom".
- This leaves the "odom"->"map" transformation to the navigation stack. This is how we want it for the demos and most use-cases.
- Previous value was only good for visualisation in RViz2 without the stack.

Note that the desired solution is specified here: https://github.com/RobotecAI/o3de-ros2-gem/issues/22

### DIFF
--- a/Code/Source/Frame/ROS2FrameComponent.cpp
+++ b/Code/Source/Frame/ROS2FrameComponent.cpp
@@ -54,7 +54,7 @@ namespace ROS2
     const char* ROS2FrameComponent::GetGlobalFrameName()
     {
         // TODO - parametrize this (typically: "odom", "world" and sometimes "map")
-        return "odom";
+        return "odom"; // Odometry frame: https://www.ros.org/reps/rep-0105.html#odom
     }
 
     bool ROS2FrameComponent::IsTopLevel() const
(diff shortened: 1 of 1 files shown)


lumberyard-employee-dm (approved):
What does "odom" stand for?

adamdbrw:
https://www.ros.org/reps/rep-0105.html#odom - the best source to learn about frames.
In our case, the navigation, we have the following considerations:
1) Localization from the SLAM stack (outside of sim) is responsible for finding out a transformation from "map" to "odom".
2) Normally, the "odom" frame relationship to "base_link" frame (the primary robot frame) would be determined by IMU (or multiple sources of odometry).
3) We decided to simplify the problem for a demo since we happen to have a ground truth about this transformation (odom->base_link). We know the location of robot perfectly well in relationship to environment's origin (0,0,0).
4) Since we have an IMU sensor, we could use its published data to also run an odometry node which would be responsible for the odom->base link transformation (and have some drift somewhat similar to real situation). This is something we could possibly do for the next Milestone.

adamdbrw:
Note: I added a comment with this REF link in the last commit.

PR #110, opened 2022-07-17
Map component

pijaro:
Map component:
- [x] MapManager component (core)
    - [x] MapReqests bus
    - [x] Map handle
    - [x] Move Geodetic transforms from gnss sensor
- [x] ROS2MapManagerComponent (ros2 comm interface)


adamdbrw:
Fixes o3de/o3de-extras#157

Code/Source/Map/MapBus.h
```diff
@@ -0,0 +1,69 @@
...
+
+        //! Convert Transform to the map coordinate system.
+        //! @param transform - Transform to be converted.
+        //! @return Copy of a Transform in the map coordinate system.
+        virtual AZ::Transform ConvertToMapCoordinateSystem(AZ::Transform transform) = 0;
+
+        //! Convert Transform from the map coordinate system.
+        //! @param transform - Transform to be converted.
+        //! @return Copy of a Transform in the world coordinate system.
+        virtual AZ::Transform ConvertFromMapCoordinateSystem(AZ::Transform transform) = 0;
```
pawellech1, line 32: here and above, if transforms will not be changed in the overriding functions it will be better to use `const AZ::Transform &` as arguments

Code/Source/Map/GeodeticConfiguration.h
```diff
@@ -0,0 +1,43 @@
...
+{
+public:
+    AZ_TYPE_INFO(GeodeticConfiguration, "{094e5059-698a-42ad-ae2d-36be20868004}");
+
+    static void Reflect(AZ::ReflectContext* context);
+
+    //! Updates the hook Transform
+    void SetHook();
+
+    bool m_useMapHook = false;
```
pawellech1, line 30: I think members of this class should be made private.

Code/Source/Map/GeodeticConfiguration.h
```diff
@@ -0,0 +1,43 @@
...
+#include <AzCore/Math/Transform.h>
+#include <AzCore/Component/Entity.h>
+#include <AzCore/std/string/string.h>
+
+namespace ROS2::Map
+{
+//! Configuration for handling geodetic map info.
+//! This configuration allows setting a geographic coordinate system hook to a simulation scene.
+//! System hook points to the real-world location in latitude, longitude, and altitude offsets.
+class GeodeticConfiguration
```
pawellech1, line 20: I don't see any constructor defined for this class. Is it omitted on purpose?

PR #114, opened 2022-07-20
Camera sensor thread safety

piotr-zyskowski-rai:
**General description**
Change to make the camera code thread-safe

**Changes in the code**
This code:
- changes readback data returned from whole readback structure to only data buffer
- ensures synchronous callback calls 
- ensures that before the CameraSensor object is destroyed, all captures were finished
  - this is done with the assumption, that for each successfully requested image, readback callback will be executed
  - this might a bold statement - needs to be fully verified
  - from quick tests it seems to work well

** Context **
Frame capture readback callback is called asynchronously so some degree of synchronization was required

### DIFF
--- a/.github/workflows/CodeChecks.yml
+++ b/.github/workflows/CodeChecks.yml
@@ -1,0 +1,45 @@
+name: CodeChecks
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    branches:
+      - development
+
+permissions:
+  contents: read
+
+jobs:
+  clang-format:
+    name: Check code formatting with clang-format-13
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone the Gem repository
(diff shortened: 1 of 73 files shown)


lgleim:
@piotr-zyskowski-rai There is a PR in O3DE upstream that is most likely relevant to the work you are doing here: https://github.com/o3de/o3de/pull/10592

piotr-zyskowski-rai:
> @piotr-zyskowski-rai There is a PR in O3DE upstream that is most likely relevant to the work you are doing here: [o3de/o3de#10592](https://github.com/o3de/o3de/pull/10592)

Thanks a lot! 
At the moment, I do not use the capture frame confirmation callback, however. As far as I understood, it does not help with the synchronization issue.
Overall, though, it's a very appreciated feature that will significantly improve API usage. :+1:

PR #120, opened 2022-07-25
Urdf importer inertial

adamdbrw:
Merge #118 first

### DIFF
--- a/Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
+++ b/Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
@@ -7,9 +7,11 @@
  */
 
 #include "URDF/RobotImporter/URDFPrefabMaker.h"
-#include "URDF/RobotImporter/TypeConversions.h"
 #include "Frame/ROS2FrameComponent.h"
+#include "URDF/RobotImporter/TypeConversions.h"
 
+#include <AtomLyIntegration/CommonFeatures/Mesh/MeshComponentBus.h>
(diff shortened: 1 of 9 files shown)


piotr-zyskowski-rai (approved):

Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
```diff
@@ -174,36 +177,100 @@ namespace ROS2
...
+        if (!inertial)
+        {   // it is ok not to have inertia in a link
+            return;
+        }
+
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
+        // TODO - consider explicit 2 arg constructor instead
+        PhysX::EditorRigidBodyConfiguration rigidBodyConfiguration;
+        rigidBodyConfiguration.m_mass = inertial->mass;
+        // TODO - is the origin.rotation part applicable? Does non-zero make value sense? Investigate.
```
pijaro, line 264: Answers to your questions: 

> is the origin.rotation part applicable?

No

> Does non-zero make value sense? 

Yes

pijaro (approved):
lgtm

Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
adamdbrw: I meant non zero for origin.rotation - which seems it doesn't make sense, not being applicable

PR #121, opened 2022-07-26
Urdf importer

adamdbrw:


Code/CMakeLists.txt
```diff
@@ -87,6 +88,9 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
         BUILD_DEPENDENCIES
             PUBLIC
                 AZ::AzToolsFramework
+                Gem::AtomLyIntegration_CommonFeatures.Editor.Static
+                Gem::LmbrCentral.Editor
```
lemonade-dm: A GEM_MODULE shouldn't be directly added as a dependency.
Instead the `LmbrCentral.API` target should be added as dependency as it is and INTERFACE target that provides the include directories for LmbrCentral.

Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -158,4 +173,13 @@ namespace ROS2
     {
         required.push_back(AZ_CRC("TransformService"));
     }
+
+    ROS2FrameComponent::ROS2FrameComponent()
+    {
+    }
```
lemonade-dm: FYI you can default this constructor here
```suggestion
    ROS2FrameComponent::ROS2FrameComponent() = default;
```

Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -158,4 +173,13 @@ namespace ROS2
...
+
+    ROS2FrameComponent::ROS2FrameComponent(const AZStd::string& frameId)
+        : m_frameName(frameId)
+    {
+    }
```
lemonade-dm: This should accept the frameId parameter  by value and move it into the m_frameName to prevent a dynamic allocation of an `AZStd::string` when a string literal is created, followed by a copy operation to copy the `AZStd::string` to member variable.

```suggestion
    ROS2FrameComponent::ROS2FrameComponent(AZStd::string frameId)
        : m_frameName(AZStd::move(frameId))
    {
    }
```

Code/Source/RobotControl/ROS2RobotControlComponent.h
```diff
@@ -21,6 +21,13 @@ namespace ROS2
     {
     public:
         AZ_COMPONENT(ROS2RobotControlComponent, "{CBFB0764-99F9-40EE-9FEE-F5F5A66E59D2}", AZ::Component);
+        ROS2RobotControlComponent()
+        {
+        }
+        ROS2RobotControlComponent(ControlConfiguration controlConfiguration)
+            : m_controlConfiguration(controlConfiguration)
```
lemonade-dm: The Controlconfiguration object can be moved here.
```suggestion
            : m_controlConfiguration(AZStd::move(controlConfiguration))
```

Code/Source/RobotImporter/RobotImporterWidgetUtils.cpp
```diff
@@ -0,0 +1,109 @@
...
+        if (!QFile::exists(path.value()))
+        {
+            QMessageBox::critical(parent, QObject::tr("Does not exist"), QObject::tr("Provided path does not exist. Please try again"));
+            return QueryUserForURDFPath();
+        }
+
+        return path->toStdString().c_str();
+    }
+
+    AZStd::optional<AZStd::string> ValidatePrefabPathExistenceAndQueryUserForNewIfNecessary(const AZStd::string& path, QWidget* parent)
```
lemonade-dm: This can accept an AZ::IO::Path and return an AZ::IO::Path since it is made for path manipulation.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+#include <SceneAPI/SceneCore/Events/AssetImportRequest.h>
+#include <SceneAPI/SceneCore/Events/SceneSerializationBus.h>
+#include <SceneAPI/SceneCore/Utilities/SceneGraphSelector.h>
+#include <Source/EditorColliderComponent.h>
+#include <Source/EditorShapeColliderComponent.h>
+
+namespace ROS2
+{
+    CollidersMaker::CollidersMaker(const AZStd::string& modelPath)
+        : m_modelPath(modelPath)
```
lemonade-dm: This should an AZStd::string by value and move it into `m_modelPath`
```suggestion
    CollidersMaker::CollidersMaker(AZStd::string modelPath)
        : m_modelPath(AZStd::move(modelPath))
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+    {
+    }
+
+    CollidersMaker::~CollidersMaker()
+    {
+    }
```
lemonade-dm: This destructor can be defaulted here.
```suggestion
    CollidersMaker::~CollidersMaker() = default;
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    void CollidersMaker::BuildColliders(urdf::LinkSharedPtr link)
+    {
+        for (auto collider : link->collision_array)
+        {
+            BuildCollider(collider);
+        }
+
+        if (link->collision_array.size() == 0)
```
lemonade-dm: Using `.empty()` better documents the intent, that a new entry shuld be added if the collision array is empty.
```suggestion
        if (link->collision_array.empty())
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+                AZ_Error("CollisionMaker", false, "Invalid json file: %s", assetInfoFilePath.c_str());
+                return;
+            }
+            auto valuesArray = valuesIterator->value.GetArray();
+            for (auto& value : valuesArray)
+            {
+                auto object = value.GetObject();
+
+                auto physXMeshGroupIterator = object.FindMember("$type");
+                AZStd::basic_string<char> physXMeshGroupType = "{5B03C8E6-8CEE-4DA0-A7FA-CD88689DD45B} MeshGroup";
```
lemonade-dm: Why not just use an `AZStd::string` here?
```suggestion
                AZStd::string physXMeshGroupType = "{5B03C8E6-8CEE-4DA0-A7FA-CD88689DD45B} MeshGroup";
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+            {
+                AZ_Error("CollisionMaker", false, "Could not save %s", assetInfoFilePath.c_str());
+                return;
+            }
+
+            // Add asset to expected assets list
+            if (assetFound)
+            {
+                auto pxmeshAssetPath = AZStd::string(assetInfo.m_relativePath.c_str());
+                AzFramework::StringFunc::Path::ReplaceExtension(pxmeshAssetPath, ".pxmesh");
```
lemonade-dm: AZ::IO::FixedMaxPath can be used for path manipulation.
```suggestion
                auto pxmeshAssetPath = AZ::IO::Path(assetInfo.m_relativePath).ReplaceExtension("pxmesh");
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+    {
+        AZStd::string typeString = "collider";
+        size_t nameSuffixIndex = 0; // For disambiguation when multiple unnamed colliders are present. The order does not matter here
+        for (auto collider : link->collision_array)
+        { // one or more colliders - the array is used
+            AddCollider(collider, entityId, PrefabMakerUtils::MakeEntityName(link->name.c_str(), typeString, nameSuffixIndex));
+            nameSuffixIndex++;
+        }
+
+        if (link->collision_array.size() == 0)
```
lemonade-dm: repeat: `.empty()` can be used to better clarify intent.
```suggestion
        if (link->collision_array.empty())
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,309 @@
...
+    {
+        AZStd::string azPath{ modelPath };
+        AZ::IO::BasicPath modelBasicPath(modelPath);
+        modelBasicPath = modelBasicPath.RemoveFilename();
+
+        AZStd::string azMeshPath{ meshPath };
+        AZ::StringFunc::Replace(azMeshPath, "package://", "", true, true);
+        modelBasicPath.Append(azMeshPath);
+
+        return modelBasicPath.String();
```
lemonade-dm: Why accept an `AZStd::basic_string<char>` while an `AZStd::string` type alias can be used.
Also it would be better to accept both parameters as an AZ::I::Path and to return an AZ::IO::Path

```suggestion
    AZ::IO::Path CollidersMaker::GetFullURDFMeshPath(AZ::IO::Path modelPath, AZ::IO::Path meshPath)
    {
        modelPath.RemoveFilename();

        AZ::StringFunc::Replace(meshPath.Native(), "package://", "", true, true);
        modelBasicPath /= meshPath;

        return modelBasicPath;
```

Code/Source/RobotImporter/URDF/CollidersMaker.h
```diff
@@ -0,0 +1,43 @@
...
+
+        //! Builds .pxmeshes for every collider in link collider mesh.
+        //! @param link A parsed URDF tree link node which could hold information about colliders.
+        void BuildColliders(urdf::LinkSharedPtr link);
+        //! Add zero, one or many collider elements (depending on link content).
+        //! @param link A parsed URDF tree link node which could hold information about colliders.
+        //! @param entityId A non-active entity which will be affected.
+        void AddColliders(urdf::LinkSharedPtr link, AZ::EntityId entityId);
+        AZStd::vector<AZStd::string> m_meshesToBuild;
+        AZStd::mutex m_buildMutex;
```
lemonade-dm: These variables should be private.

Also I recommended making m_meshesToBuild a vector of `AZ::IO::Path` objects.

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+            return "";
+        }
+
+        auto relativePath = std::filesystem::path(assetInfo.m_relativePath.c_str());
+        relativePath.replace_extension("azmodel");
+        auto assetPath = AZStd::string(relativePath.string().c_str());
+        AZStd::to_lower(assetPath.begin(), assetPath.end());
+
+        return assetPath;
+    }
```
lemonade-dm: I recommended using the AZ::IO::Path API here.
It has the ability to get a modifiable lvalue reference to the underlying string instance via the [Native()](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/Path/Path.h#L537-L538) function.
That can be used to populate the underlying `AZStd::string` instance stored in an AZ::IO::Path when passed to EBus functions.

```suggestion
    AZ::IO::Path PrefabMakerUtils::GetAzModelAssetPathFromModelPath(AZ::IO::Path modelPath)
    {
        bool assetFound = false;
        AZ::Data::AssetInfo assetInfo;
        AZ::IO::Path watchDir;
        AzToolsFramework::AssetSystemRequestBus::BroadcastResult(
            assetFound,
            &AzToolsFramework::AssetSystem::AssetSystemRequest::GetSourceInfoBySourcePath,
            modelPath.c_str(),
            assetInfo,
            watchDir.Native());

        if (!assetFound)
        {
            AZ_Error("AddVisuals", false, "Could not find model asset for %s", modelPath.c_str());
            return {};
        }

        auto asssetPath = AZ::IO::FixedMaxPath(assetInfo.m_relativePath).ReplaceExtension("azmodel");
        AZStd::to_lower(assetPath.Native().begin(), assetPath.Native().end());

        return assetPath;
    }
```

Code/Source/RobotImporter/URDF/PrefabMakerUtils.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "UrdfParser.h"
+#include <AzToolsFramework/Prefab/PrefabPublicInterface.h>
+
+#include <filesystem> // TODO - instead, use AZ API for filesystem
+
+namespace ROS2
+{
+    //! Common utils for Prefab Maker classes
+    class PrefabMakerUtils
```
lemonade-dm: Instead of making a class, a namespace can be used instead.

The drawback of placing static functions with a class, is that the functions can't be forward declared, if the user didn't want to include a header for some reason.

Code/Source/RobotImporter/URDF/PrefabMakerUtils.h
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include "UrdfParser.h"
+#include <AzToolsFramework/Prefab/PrefabPublicInterface.h>
+
+#include <filesystem> // TODO - instead, use AZ API for filesystem
```
lemonade-dm: Looks like this can away now with my previous comment.

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,209 @@
...
+            break;
+        case urdf::Geometry::MESH:
+            {
+                auto meshGeometry = std::dynamic_pointer_cast<urdf::Mesh>(geometry);
+
+                // TODO - a PoC solution for path, replace with something generic, robust, proper
+                std::filesystem::path modelPath(m_modelPath.c_str());
+                modelPath = modelPath.remove_filename();
+                auto relativePathToMesh = std::regex_replace(meshGeometry->filename, std::regex("package://"), "");
+                modelPath += relativePathToMesh;
```
lemonade-dm: The reason I recommended against using the `std::filesystem::path` functions is that it exposes what I believe is a dangerous [string concatenation operator](https://en.cppreference.com/w/cpp/filesystem/path/concat).

This code here is joining two paths together smartly with a path separator in between but instead appending two strings.
You would want to use the `operator/=` here, not `operator+=`.
The AZ::IO::Path API omits the plus('+') operator to avoid that danger.

```suggestion
                AZ::IO::Path modelPath(m_modelPath);
                modelPath().RemoveFilename(); 
                AZ::IO::Path meshFilename(AZStd::string_view(meshGeometry->filename.c_str(), meshGeometry->filename.size()));
                AZ::IO::StringFunc::Replace(meshFilename.Native(), "package://", "", true, true);
                modelPath /= meshFilename;
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
adamdbrw: @pijaro I am correcting this, let me know if there was a specific reason

Code/Source/RobotImporter/URDF/CollidersMaker.h
adamdbrw: They are in use outside of the class currently, but I see that it is due to a suboptimal design. URDFPrefabMaker will call a self-contained member function of CollidersMaker instead of a current solution which violates encapsulation. This will be done in the second commit.

adamdbrw:
@lumberyard-employee-dm Thank you for your review. All the suggested changes have been committed.  
I was wondering whether we could also have domain specialists in topics such as Prefab System and Asset Builder take a look at this.

lemonade-dm:
> @lumberyard-employee-dm Thank you for your review. All the suggested changes have been committed. I was wondering whether we could also have domain specialists in topics such as Prefab System and Asset Builder take a look at this.

I'll reach out and see if I can wrangle some of the Prefab and Asset Builder system engineers.

AMZN-stankowi:
Based on some of the renamed filenames, it looks like you're converting URDF to FBX and then processing FBX from there. Is there a reason you aren't adding URDF support to our 3D scene asset importer library ( https://github.com/assimp/assimp ) instead? Then it should process and load just like FBX, GLTF, STL, and other supported file formats.

adamdbrw:
@AMZN-stankowi 
This is not the case, but I understand the source of confusion. The FBX converter is there in an initial phase, but we do not use this conversion approach.
Instead, we construct a robot prefab and populate it with entities and components based on URDF tree (directly).

Our first version is an Importer tool. There are domain differeces which might not be trivial to solve, and follow up steps to do with the imported robot before it is ready (such as adding sensors and robot control). This would suggest an Importer (wizard -like) is a better fit, but extension of assimp is something we would like to look at in the next iteration. The logic of prefab construction is there to reuse.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,341 @@
...
+            AzToolsFramework::AssetSystemRequestBus::BroadcastResult(
+                assetFound,
+                &AzToolsFramework::AssetSystem::AssetSystemRequest::GetSourceInfoBySourcePath,
+                azMeshPath.c_str(),
+                assetInfo,
+                watchDir);
+
+            // We are expecting .pxmesh
+            auto pxmodelPath = AZStd::string(assetInfo.m_relativePath.c_str());
+            AzFramework::StringFunc::Path::ReplaceExtension(pxmodelPath, ".pxmesh");
```
AMZN-stankowi, line 247: This is generally a frail way to get the product asset information for a source asset. There's no guarantee that product asset filenames will be the source asset filename with an extension swap. Especially in the case of scene files, where you could have something like Shapes.fbx output Cube.pxmesh, Cylinder.pxmesh, etc, and not have any product assets with Shapes in the filename.

AMZN-stankowi: To add some more context: We've been talking a lot about asset references recently on the asset pipeline team. There are several problems in this area, but the core of it is path references to assets can be very unstable.

I've been working on updated documentation to try to make it easier to understand how to manage assets referencing other assets, you can see the pull request here that I'm hoping to get in soon https://github.com/o3de/o3de.org/pull/1882

Code/Source/ROS2EditorModule.cpp
```diff
@@ -24,9 +25,7 @@ namespace ROS2
             // EditContext. This happens through the [MyComponent]::Reflect() function.
             m_descriptors.insert(
                 m_descriptors.end(),
-                {
-                    ROS2EditorSystemComponent::CreateDescriptor(),
-                });
+                { ROS2EditorSystemComponent::CreateDescriptor(), ROS2RobotImporterEditorSystemComponent::CreateDescriptor() });
```
spham-amzn, line 28: minor: Can you format (align) the descriptor list so its easier to read the list?

Code/Source/ROS2ModuleInterface.h
```diff
@@ -46,9 +48,7 @@ namespace ROS2
         //! Add required SystemComponents to the SystemEntity.
         AZ::ComponentTypeList GetRequiredSystemComponents() const override
         {
-            return AZ::ComponentTypeList{
-                azrtti_typeid<ROS2SystemComponent>(),
-            };
+            return AZ::ComponentTypeList{ azrtti_typeid<ROS2SystemComponent>(), azrtti_typeid<ROS2RobotImporterSystemComponent>() };
```
spham-amzn, line 51: minor: Can you format and left-align the component type list here as well?

Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
```diff
@@ -0,0 +1,61 @@
...
+        options.showOnToolsToolbar = true;
+        // options.toolbarIcon = ":/.svg"; // TODO - icon
+
+        AzToolsFramework::RegisterViewPane<RobotImporterWidget>("Robot Importer", "ROS2", options);
+    }
+
+    ROS2RobotImporterEditorSystemComponent::ROS2RobotImporterEditorSystemComponent() = default;
+
+    ROS2RobotImporterEditorSystemComponent::~ROS2RobotImporterEditorSystemComponent() = default;
+} // namespace ROS2
```
spham-amzn: minor:
You could also declare them as default in the header file instead of the cpp file

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace ROS2
+{
+    namespace Internal
+    {
+        AZStd::string GetBaseName(AZStd::string path)
+        {
+            QFileInfo fileInfo(path.c_str());
+            return fileInfo.baseName().toStdString().c_str();
+        }
```
spham-amzn: There is already functionality in AzCore to do this, no need to re-create it here. Refer to Code/Framework/AzCore/AzCore/IO/Path/Path.h

```
...
class PathView 
{
...
    constexpr PathView Filename() const;
...

```
So the usage would look like this
```
AZ::IO::PathView(path).Filename().Native()
```

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -0,0 +1,114 @@
...
+                  case RobotImporter::LogLevel::Error:
+                      ReportError(message);
+                      break;
+                  }
+              })
+    {
+        setWindowTitle(QObject::tr("Robot definition file importer"));
+        QVBoxLayout* mainLayout = new QVBoxLayout(this);
+        mainLayout->setSpacing(20);
+        QLabel* captionLabel = new QLabel(QObject::tr("Select a robot definition (URDF) file to import"), this);
```
spham-amzn: Should this read
`Select a Unified Robot Description Format  (URDF) file to import` ?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -0,0 +1,114 @@
...
+            this,
+            [this]()
+            {
+                AZStd::optional<AZStd::string> urdfPath = RobotImporterWidgetUtils::QueryUserForURDFPath(this);
+                if (!urdfPath)
+                {
+                    return;
+                }
+
+                auto prefabName = AZStd::string::format("%s.%s", Internal::GetBaseName(urdfPath.value()).c_str(), "prefab");
```
spham-amzn: No need to format specify for 'prefab' if its a constant string
```suggestion
                auto prefabName = AZStd::string::format("%s.prefab", Internal::GetBaseName(urdfPath.value()).c_str());
```

lemonade-dm: Alternatively the `Internal::GetBaseName(urdfPath.value())` can be placed in an AZ::IO::Path and then the [ReplaceExtension](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/Path/Path.h#L528-L532) can be used to set prefab as the extension.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
adamdbrw: Could you, at this moment, suggest a better approach?

Code/Source/ROS2ModuleInterface.h
adamdbrw: This is actually what our clang-format settings do (I just double-checked running clang-format for the file). It should be the same format as in O3DE.
Do you think we should change the format settings?

Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
adamdbrw: see above

Code/Source/ROS2EditorModule.cpp
adamdbrw: See https://github.com/RobotecAI/o3de-ros2-gem/pull/121#discussion_r958545475

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
AMZN-stankowi: Is there where you're retrieving the FBX filename to try to extract for the physics mesh?
`auto azMeshPath = GetFullURDFMeshPath(AZ::IO::Path(m_modelPath), AZ::IO::Path(meshGeometry->filename.c_str()));`

If so, I would say either update this to point at the product asset of the physx mesh instead of the source asset of the FBX file, or if you need it for a few reasons, then add additional user data of the physx mesh to use. The URDF author is going to need to pick the correct mesh to set in the URDF file in case the FBX file has multiple physx mesh product assets, and/or the physx mesh product asset doesn't match the filename of the FBX.

adamdbrw: I believe that for URDF `<collider>` and `<visual>` tag only a file with a single mesh is allowed. Each link can have multiple  of these tags and this is how multiple colliders/visuals per part are handled.
Let me double check that.
Note that we will be typically working with STL or DAE files, not FBX.

Code/Source/ROS2ModuleInterface.h
spham-amzn: No, it was a minor comment. If this is clang-formatted, then I'd prefer its consistency

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,341 @@
...
+                AZ_Error("CollisionMaker", false, "Invalid json file: %s", assetInfoFilePath.c_str());
+                return;
+            }
+            auto valuesArray = valuesIterator->value.GetArray();
+            for (auto& value : valuesArray)
+            {
+                auto object = value.GetObject();
+
+                auto physXMeshGroupIterator = object.FindMember("$type");
+                AZStd::string physXMeshGroupType = "{5B03C8E6-8CEE-4DA0-A7FA-CD88689DD45B} MeshGroup";
```
spham-amzn: This should be a constant constexpr string declared outside of this loop

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
adamdbrw: @pijaro take a look please

adamdbrw:
@lumberyard-employee-dm  @spham-amzn All review changes suggested by you so far are committed, let me know if there is anything else you would like to see changed.
https://github.com/RobotecAI/o3de-ros2-gem/pull/121#discussion_r958568335 is something @pijaro will be looking into, we might push a fix to that. Currently, the potential problem does not seem to be affecting URDFs we tested, this might also be delegated to an issue if it is: 1) not impactful at the moment 2) a larger task.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
piotr-zyskowski-rai: I took over the task from @pijaro 

I implemented the path query using 
`AzToolsFramework::AssetSystem::AssetSystemRequest::GetAssetsProducedBySourceUUID` and `AZ::Data::AssetCatalogRequestBus::Events::GetAssetPathById` and then I filter the paths by extension searching for `.pxmesh` extension. Also, at the moment, I assume that from each `.dae` model only one `.pxmesh` is being created.

@AMZN-stankowi Is it a sensible approach?

Since it is a bit bigger change I created a separate PR (#156) for this feature.

The PR deals with extension replacement from this review thread but there is [another place](https://github.com/RobotecAI/o3de-ros2-gem/pull/121/files#diff-1d0a02d3c5cc965628b4c5487fae26cf3de67ad9d6025b0631336ca7869a2383R169) where such replacement is used. I'll look into this.

adamdbrw: I will resolve this conversation, please continue under #156

adamdbrw:
@lumberyard-employee-dm @spham-amzn @AMZN-stankowi  could you please take another look / approve?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+                m_selectFileButton.setEnabled(false);
+                // Check whether import is still in progress every 0.5 seconds
+                m_importerUpdateTimer.start(500);
+            });
+        setLayout(mainLayout);
+    }
+
+    void RobotImporterWidget::ReportError(const AZStd::string& errorMessage)
+    {
+        QMessageBox::critical(this, QObject::tr("Error"), QObject::tr(errorMessage.c_str()));
```
spham-amzn, line 96: Is it a good idea to trigger a message box here during an import? What if there was a problem that affected a lot of assets and triggers an error. This message box could pop up a lot of times, forcing the operator to click ok many times.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,341 @@
...
+        {
+            auto meshGeometry = std::dynamic_pointer_cast<urdf::Mesh>(geometry);
+            auto azMeshPath = GetFullURDFMeshPath(AZ::IO::Path(m_modelPath), AZ::IO::Path(meshGeometry->filename.c_str()));
+
+            AZStd::shared_ptr<AZ::SceneAPI::Containers::Scene> scene;
+            AZ::SceneAPI::Events::SceneSerializationBus::BroadcastResult(
+                scene, &AZ::SceneAPI::Events::SceneSerialization::LoadScene, azMeshPath.c_str(), AZ::Uuid::CreateNull());
+            if (!scene)
+            {
+                AZ_Error("CollisionMaker", false, "Error loading collider. Invalid scene.");
```
spham-amzn: Is there any ID or name in the collision object that can be logged here as part of the message?

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,341 @@
...
+            AZ::SceneAPI::Events::AssetImportRequestBus::BroadcastResult(
+                result,
+                &AZ::SceneAPI::Events::AssetImportRequest::UpdateManifest,
+                *scene,
+                AZ::SceneAPI::Events::AssetImportRequest::ManifestAction::Update,
+                AZ::SceneAPI::Events::AssetImportRequest::RequestingApplication::Editor);
+
+            if (result.GetResult() != AZ::SceneAPI::Events::ProcessingResult::Success)
+            {
+                AZ_Error("CollisionMaker", false, "Scene updated");
```
spham-amzn: "Scene updated" is an error?

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,341 @@
...
+                &AzToolsFramework::AssetSystem::AssetSystemRequest::GetSourceInfoBySourcePath,
+                azMeshPath.c_str(),
+                assetInfo,
+                watchDir);
+
+            // Set export method to convex mesh
+            auto readOutcome = AZ::JsonSerializationUtils::ReadJsonFile(assetInfoFilePath.c_str());
+            if (!readOutcome.IsSuccess())
+            {
+                AZ_Error("CollisionMaker", false, "Could not read %s", assetInfoFilePath.c_str());
```
spham-amzn: Should include the error from the readOutcome in the error message as well
```readOutcome.GetError().c_str()```

File: Code/Source/RobotImporter/URDF/CollidersMaker.cpp [hunk 30]
```diff
@@ -0,0 +1,341 @@
...
+                return;
+            }
+            rapidjson::Document assetInfoJson = readOutcome.TakeValue();
+            auto manifestObject = assetInfoJson.GetObject();
+            auto valuesIterator = manifestObject.FindMember("values");
+            if (valuesIterator == manifestObject.MemberEnd())
+            {
+                AZ_Error("CollisionMaker", false, "Invalid json file: %s", assetInfoFilePath.c_str());
```
spham-amzn: Should add a message indication thaat 'values' was not found.
```suggestion
                AZ_Error("CollisionMaker", false, "Invalid json file: %s (Missing 'values' node)", assetInfoFilePath.c_str());
```

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,341 @@
...
+                if (AZ::StringFunc::Equal(physXMeshGroupIterator->value.GetString(), physXMeshGroupType))
+                {
+                    value.AddMember(rapidjson::StringRef("export method"), rapidjson::StringRef("1"), assetInfoJson.GetAllocator());
+                }
+            }
+
+            auto saveOutcome = AZ::JsonSerializationUtils::WriteJsonFile(assetInfoJson, assetInfoFilePath.c_str());
+            if (!saveOutcome.IsSuccess())
+            {
+                AZ_Error("CollisionMaker", false, "Could not save %s", assetInfoFilePath.c_str());
```
spham-amzn: Same here, add the error from saveOutcome to the error message
```saveOutcome.GetError().c_str()```

File: Code/Source/RobotImporter/URDF/CollidersMaker.cpp [hunk 32]
```diff
@@ -0,0 +1,341 @@
...
+    {
+        AZStd::string typeString = "collider";
+        size_t nameSuffixIndex = 0; // For disambiguation when multiple unnamed colliders are present. The order does not matter here
+        for (auto collider : link->collision_array)
+        { // one or more colliders - the array is used
+            AddCollider(collider, entityId, PrefabMakerUtils::MakeEntityName(link->name.c_str(), typeString, nameSuffixIndex));
+            nameSuffixIndex++;
+        }
+
+        if (link->collision_array.empty())
```
spham-amzn: nit: you could also check if nameSuffixIndex==0 here instead :-)

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Multiple errors during import are only logged into the console/log file.

`ReportError` function is intended to be used as a final message to the user (that the import failed and why). It should not be called repeatedly within the importer pipeline.

adamdbrw: @spham-amzn 
This is a valid concern for sure.
As @piotr-zyskowski-rai mentioned, this box only handles the end result. However..
URDF importer in current version has very little UX.
Considering that it is already a huge PR, I propose to handle UX in a dedicated issue: https://github.com/RobotecAI/o3de-ros2-gem/issues/163.
Let me know if this is ok.

adamdbrw:
@michalpelka will submit fixes following @spham-amzn review comments today.

Code/Source/RobotImporter/RobotImporterWidget.cpp
spham-amzn: As long as this is triggered at the end of the entire process, then its fine.

michalpelka:
All your comments @spham-amzn were addressed - thanks for reviewing.
I would be happy if you take a look and comment / approve.

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+        AzToolsFramework::AssetSystemRequestBus::BroadcastResult(
+            assetFound,
+            &AzToolsFramework::AssetSystem::AssetSystemRequest::GetSourceInfoBySourcePath,
+            modelPath.c_str(),
+            assetInfo,
+            watchDir.Native());
+
+        if (!assetFound)
+        {
+            AZ_Error("AddVisuals", false, "Could not find model asset for %s", modelPath.c_str());
```
spham-amzn: is 'AddVisuals' the correct log name here?

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+        AZ::EntityId entityId = createEntityResult.GetValue();
+        if (!entityId.IsValid())
+        { // Verify that a valid entity is created.
+            return AZ::Failure(AZStd::string("Invalid id for created entity"));
+        }
+
+        AZ_TracePrintf("CreateEntity", "Processing entity id:%s with name:%s", entityId.ToString().c_str(), name.c_str());
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
+        AZStd::string entityName(name.c_str());
```
spham-amzn: This entityName local variable seems unnecessary, why not just pass in name into entity->SetName()?

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    AzToolsFramework::Prefab::PrefabOperationResult RemoveEntityWithDescendants(AZ::EntityId parentEntityId)
+    {
+        auto prefabInterface = AZ::Interface<AzToolsFramework::Prefab::PrefabPublicInterface>::Get();
+        return prefabInterface->DeleteEntitiesAndAllDescendantsInInstance({ parentEntityId });
+    }
+
+    void AddRequiredComponentsToEntity(AZ::EntityId entityId)
+    {
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
```
spham-amzn, line 99: Looks like a good place to protect against an invalid ```entityId```. (```*entity``` below may cause a NPE)

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+        AzToolsFramework::EditorEntityContextRequestBus::Broadcast(
+            &AzToolsFramework::EditorEntityContextRequests::AddRequiredComponents, *entity);
+    }
+
+    bool HasCollider(AZ::EntityId entityId)
+    { // TODO - actually, we want EditorColliderComponent specifically, but the change can be applied only after moving to ECC
+        // TODO - which will happen when Cylinder shape is supported. Until then, we check for either ECC or ESCC.
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
```
spham-amzn, line 108: Same comment here, (or at least have an assert on ```entity```

Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <PhysX/MeshAsset.h>
+
+#include <regex> // TODO - we are currently replacing package:// with an absolute path
+
+namespace ROS2
+{
+    URDFPrefabMaker::URDFPrefabMaker(const AZStd::string& modelFilePath, urdf::ModelInterfaceSharedPtr model, AZStd::string prefabPath)
+        : m_model(model)
+        , m_visualsMaker(modelFilePath, model->materials_)
+        , m_collidersMaker(new CollidersMaker(modelFilePath))
```
spham-amzn: maybe make m_collidersMaker a AZStd::unique_ptr instead of a shared_ptr? And use AZStd::make_unique<> instead of new

Code/Source/RobotControl/ROS2RobotControlComponent.h
```diff
@@ -21,6 +21,13 @@ namespace ROS2
     {
     public:
         AZ_COMPONENT(ROS2RobotControlComponent, "{CBFB0764-99F9-40EE-9FEE-F5F5A66E59D2}", AZ::Component);
+        ROS2RobotControlComponent()
+        {
+        }
```
lemonade-dm: This can be "=" default'ed
```suggestion
        ROS2RobotControlComponent() = default;
```

Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.h
```diff
@@ -0,0 +1,36 @@
...
+
+        ROS2RobotImporterEditorSystemComponent() = default;
+        ~ROS2RobotImporterEditorSystemComponent() = default;
+
+    private:
+        void Activate() override;
+        void Deactivate() override;
+        void NotifyRegisterViews() override;
+    };
+} // namespace ROS2
```
lemonade-dm: nitpick: missing newline at the end of this file.

Code/Source/RobotImporter/RobotImporterWidgetUtils.cpp
```diff
@@ -0,0 +1,109 @@
...
+        }
+    } // namespace Internal
+
+    AZStd::optional<AZStd::string> QueryUserForURDFPath(QWidget* parent)
+    {
+        std::optional<QString> path = Internal::QueryUserForPath("Unified Robot Description Format (*.urdf)", QFileDialog::ExistingFiles);
+        if (!path)
+        {
+            return AZStd::nullopt;
+        }
```
lemonade-dm: nitpick: AZStd::option can be used here(It is just an alias of std::optional)

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <LmbrCentral/Shape/EditorShapeComponentBus.h>
+#include <LmbrCentral/Shape/SphereShapeComponentBus.h>
+
+namespace ROS2
+{
+    VisualsMaker::VisualsMaker(const AZStd::string& modelPath, const std::map<std::string, urdf::MaterialSharedPtr>& materials)
+        : m_modelPath(modelPath)
+        , m_materials(materials)
+    {
+    }
```
spham-amzn, line 38: Should assert on an empty m_modelPath (like in URDFPrefabMaker::URDFPrefabMaker)

Code/Source/RobotImporter/RobotImporterWidgetUtils.cpp
```diff
@@ -0,0 +1,109 @@
...
+            return QueryUserForURDFPath();
+        }
+
+        if (!QFile::exists(path.value()))
+        {
+            QMessageBox::critical(parent, QObject::tr("Does not exist"), QObject::tr("Provided path does not exist. Please try again"));
+            return QueryUserForURDFPath();
+        }
+
+        return path->toStdString().c_str();
```
lemonade-dm: `path.toUtf8().constData()` allows for use of a path with UTF-8 characters on it and avoids the need to allocate a std::string. In this scenario a QByteArray is created

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    void VisualsMaker::AddVisuals(urdf::LinkSharedPtr link, AZ::EntityId entityId)
+    {
+        const AZStd::string typeString = "visual";
+        if (link->visual_array.size() == 0)
+        { // one or zero visuals - element is used
```
spham-amzn, line 44: This comment doesnt match the condition "one or zero visuals". Should the size check for 1 or 0?

Code/Source/RobotImporter/RobotImporterWidgetUtils.cpp
```diff
@@ -0,0 +1,109 @@
...
+            // TODO: It should be fixed at some point in the future
+            AZStd::optional<QString> newPathCandidate = Internal::QueryUserForPath("Prefab (*.prefab)", QFileDialog::AnyFile);
+            if (!newPathCandidate || newPathCandidate->isEmpty())
+            {
+                return AZStd::nullopt;
+            }
+            return ValidatePrefabPathExistenceAndQueryUserForNewIfNecessary(newPathCandidate.value().toStdString().c_str());
+        }
+    }
+} // namespace ROS2::RobotImporterWidgetUtils
```
lemonade-dm: nitpick missing newline at the end of this file.

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+                if (!urdfPath)
+                {
+                    return;
+                }
+
+                AZ::IO::Path prefabName(AZ::IO::PathView(urdfPath.value()).Filename());
+                prefabName.ReplaceExtension("prefab");
+                AZStd::string prefabDefaultPath(AZ::IO::Path(AZ::Utils::GetProjectPath()) / "Assets" / "Importer" / prefabName);
+                auto prefabPath =
+                    RobotImporterWidgetUtils::ValidatePrefabPathExistenceAndQueryUserForNewIfNecessary(prefabDefaultPath, this);
```
lemonade-dm: FYI: you can get an lvalue `AZStd::string&` from an AZ::IO::Path via using the `Native()` method.

That allows interoperability functions that take an AZStd::string or AZStd::fixed_string<1024>

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // Apply transform as per origin
+        PrefabMakerUtils::SetEntityTransform(visual->origin, entityId);
+
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
+        auto geometry = visual->geometry;
+        switch (geometry->type)
+        {
+        case urdf::Geometry::SPHERE:
+            {
+                auto sphereGeometry = std::dynamic_pointer_cast<urdf::Sphere>(geometry);
```
spham-amzn, line 97: RTTI is disabled by default on the global compile settings. Modules that need RTTI will need to explicitly enable it. Is this relying on RTTI? If so, there doesn't seem to be a nullptr check for this object

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,346 @@
...
+                        AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(50));
+                    }
+                }
+
+                AZ_Printf("CollisionMaker", "All URDF assets ready!");
+                // Notify the caller that we can continue with constructing the prefab.
+                notifyBuildReadyCb();
+            });
+    }
+} // namespace ROS2
```
lemonade-dm: nitpick: missing newline at the end of the file

spham-amzn:
External comment, there are a lot of ```TODO``` comments here. Are we tracking things that need to be done anywhere besides in these comments?

Code/Source/RobotImporter/URDF/CollidersMaker.h
```diff
@@ -0,0 +1,54 @@
...
+#include <AzCore/Component/EntityId.h>
+#include <AzCore/IO/Path/Path.h>
+#include <AzCore/std/containers/vector.h>
+#include <AzCore/std/parallel/atomic.h>
+#include <AzCore/std/parallel/mutex.h>
+#include <AzCore/std/parallel/thread.h>
+
+namespace ROS2
+{
+    typedef std::function<void()> BuildReadyCallback;
```
lemonade-dm: The AZStd library has a std::function implementation(that also supports a small buffer optimization of at least 16 bytes to avoid memory allocations for functions objects smaller than that).

```suggestion
    using BuildReadyCallback = std::function<void()>;
```

Also using is preferred to typedef, as it supports template aliases.
Also subset developers believe it is more readable for function pointer declarations
i.e `typedef void(*callback)()` vs `using callback = void(*)()`

Code/Source/RobotImporter/URDF/CollidersMaker.h
```diff
@@ -0,0 +1,54 @@
...
+        AZ::IO::Path GetFullURDFMeshPath(AZ::IO::Path modelPath, AZ::IO::Path meshPath);
+
+        AZStd::string m_modelPath;
+
+        AZStd::thread m_buildThread;
+        AZStd::mutex m_buildMutex;
+        AZStd::vector<AZ::IO::Path> m_meshesToBuild;
+        AZStd::atomic_bool m_stopBuildFlag;
+    };
+} // namespace ROS2
```
lemonade-dm: nitpick: missing newline at the end of the file

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <AzCore/std/string/conversions.h>
+#include <AzToolsFramework/API/EditorAssetSystemAPI.h>
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+#include <AzToolsFramework/ToolsComponents/TransformComponent.h>
+#include <Source/EditorColliderComponent.h>
+#include <Source/EditorShapeColliderComponent.h>
+
+namespace ROS2::PrefabMakerUtils
+{
+    AZ::IO::Path GetAzModelAssetPathFromModelPath(AZ::IO::Path modelPath)
```
lemonade-dm: This could accept the AZ::IO::Path by const reference since it is not modified.

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+    AzToolsFramework::EntityIdList GetColliderChildren(AZ::EntityId parentEntityId)
+    {
+        AzToolsFramework::EntityIdList colliderChildren;
+        AzToolsFramework::EntityIdList allChildren = AzToolsFramework::GetEntityChildOrder(parentEntityId);
+        for (auto childId : allChildren)
+        {
+            AZ_TracePrintf("GetColliderChildren", "Considering child %s", childId.ToString().c_str());
+            if (HasCollider(childId))
+            {
+                AZ_TracePrintf("GetColliderChildren", "Child %s has a collider", childId.ToString().c_str());
```
lemonade-dm: One thing that isn't known by users of `AZ_TracePrintf` is that unlike `AZ_Error`, `AZ_Warning` and `AZ_Assert` it doesn't automatically append a newline to the end of the message.

So the all the trace prtinfs will be output on a single line.

That being said, are these traceprints meant to stick around, since they seem like debug printfs to test this function and not to log useful user facing information.

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -0,0 +1,134 @@
...
+                colliderChildren.push_back(childId);
+            }
+        }
+        return colliderChildren;
+    }
+
+    AZStd::string MakeEntityName(const AZStd::string& rootName, const AZStd::string& type, size_t index)
+    {
+        const AZStd::string suffix = index == 0 ? AZStd::string("") : AZStd::string::format("_%zu", index);
```
lemonade-dm: Actually it is better to use the AZStd::string default constructor `AZStd::string()` over the AZStd::string c-string constructor `AZStd::string("")` as the c-string constructor must invoke strlen on an empty string to check that it has a size of 0,.

The default constructor doesn't need to do this, so it is slightly faster.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,346 @@
...
+                            }
+                        }
+                    }
+                    if (!m_meshesToBuild.empty())
+                    {
+                        AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(50));
+                    }
+                }
+
+                AZ_Printf("CollisionMaker", "All URDF assets ready!");
```
lemonade-dm: `AZ_Printf` doesn't automatically add newlines, so you probably want to add newlines to end of each AZ_Printf.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -0,0 +1,346 @@
...
+                            AZ::Data::AssetId assetId;
+                            AZ::Data::AssetType assetType = AZ::AzTypeInfo<PhysX::Pipeline::MeshAsset>::Uuid();
+                            AZ::Data::AssetCatalogRequestBus::BroadcastResult(
+                                assetId, &AZ::Data::AssetCatalogRequests::GetAssetIdByPath, iter->Native().c_str(), assetType, false);
+                            if (assetId.IsValid())
+                            {
+                                AZ_Printf("CollisionMaker", "Asset %s found and valid...", iter->Native().c_str());
+                                m_meshesToBuild.erase(iter--);
+                            }
+                        }
```
lemonade-dm: You can use the [AZStd::erase_if](https://en.cppreference.com/w/cpp/container/vector/erase2) function to remove the paths from the `m_meshesToBuild` container.

nitpick: The AZ::IO::Path class has a `.c_str()` function on it which invokes c_str on the underlying string type.
So `.Native().c_str()` can be replace with just `c_str()`

```suggestion
                        auto eraseFoundMesh = [](const AZ::IO::Path& meshPath)
                        {
                            AZ::Data::AssetId assetId;
                            AZ::Data::AssetType assetType = AZ::AzTypeInfo<PhysX::Pipeline::MeshAsset>::Uuid();
                            AZ::Data::AssetCatalogRequestBus::BroadcastResult(
                                assetId, &AZ::Data::AssetCatalogRequests::GetAssetIdByPath, meshPath.c_str(), assetType, false);
                            if (assetId.IsValid())
                            {
                                AZ_Printf("CollisionMaker", "Asset %s found and valid...", meshPath.c_str());
                                return true; // return true to erase the mesh path
                            }
                            
                            return false;
                        };

                        AZStd::erase_if(m_meshesToBuild, AZStd::move(eraseFoundMesh ));
```

Code/Source/RobotImporter/URDF/RobotImporter.h
```diff
@@ -0,0 +1,58 @@
...
+        //! Checks if assets are loaded and if so, proceeds with prefab creation
+        //! This function should be called from the main thread repeatedly after ParseURDFAndStartLoadingAssets
+        //! It will poll the asset import process and if it is finished, it will proceed with prefab creation
+        //! @param importFinishedCb Function that is called when the import process is finished
+        void CheckIfAssetsWereLoadedAndCreatePrefab(std::function<void()> importFinishedCb);
+
+    private:
+        std::atomic_bool m_isProcessingAssets;
+        std::optional<URDFPrefabMaker> m_prefabMaker;
+        std::function<void(LogLevel, const AZStd::string&)> m_logger;
```
lemonade-dm: Seems like this could use AZStd:: types instead of std::.

Also it may help to better document the Logger function callback by making a type alias for it.
`using LoggerFunction = AZStd::function<void(LogLevel, AZStd::string_view)>;`

Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+#include "RobotImporter/RobotImporterWidget.h"
+
+#include <AzCore/IO/FileIO.h>
+#include <AzCore/Utils/Utils.h>
+#include <AzToolsFramework/API/EditorAssetSystemAPI.h>
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+#include <PhysX/MeshAsset.h>
+
+#include <regex> // TODO - we are currently replacing package:// with an absolute path
```
lemonade-dm: This regex include isn't used(Also there is an AzCore/std/string/regex.h implementation as well)

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,208 @@
...
+            LmbrCentral::EditorShapeComponentRequestsBus::Event(
+                entityId, &LmbrCentral::EditorShapeComponentRequests::SetShapeColor, materialColor);
+            entity->Deactivate();
+            return;
+        }
+
+        // Mesh visual - we can have either filename or default material with a given color
+        // TODO - handle texture_filename - file materials
+        entity->CreateComponent(AZ::Render::EditorMaterialComponentTypeId);
+        AZ_Warning("AddVisual", false, "Setting color for material %s", visual->material->name.c_str());
```
lemonade-dm: Is this really a warning or a just a regular info log?

For a regular information log, `AZ_TracePrintf` should be used with a newline at the end.

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,208 @@
...
+            // Material is optional, and it requires geometry
+            return;
+        }
+
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
+
+        // If present in map, take map color definition as priority, otherwise apply local node definition
+        urdf::Color materialColorUrdf = m_materials.find(visual->material->name) == m_materials.end()
+            ? visual->material->color
+            : m_materials[visual->material->name]->color;
```
lemonade-dm: If you switch the m_materials type to use `AZStd::map`, this can use the C++20 [contains](https://en.cppreference.com/w/cpp/container/map/contains) function

Code/Source/RobotImporter/URDF/VisualsMaker.h
```diff
@@ -0,0 +1,36 @@
...
+        //! @param entityId A non-active entity which will be affected.
+        void AddVisuals(urdf::LinkSharedPtr link, AZ::EntityId entityId);
+
+    private:
+        void AddVisual(urdf::VisualSharedPtr visual, AZ::EntityId entityId, const AZStd::string& generatedName);
+        void AddVisualToEntity(urdf::VisualSharedPtr visual, AZ::EntityId entityId);
+        void AddMaterialForVisual(urdf::VisualSharedPtr visual, AZ::EntityId entityId);
+
+        AZStd::string m_modelPath; // TODO - this should be handled on level of assets
+        std::map<std::string, urdf::MaterialSharedPtr> m_materials;
```
lemonade-dm: The `m_modelPath` can be made an AZ::IO::Path and the m_materials can be an `AZStd::map` or `AZStd::unordered_map` if order isn't important.

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        QMessageBox::critical(this, QObject::tr("Error"), QObject::tr(errorMessage.c_str()));
+        m_statusLabel.setText(errorMessage.c_str());
+        AZ_Error("RobotImporterWidget", false, errorMessage.c_str());
+    }
+
+    void RobotImporterWidget::ReportInfo(const AZStd::string& infoMessage)
+    {
+        m_statusLabel.setText(QObject::tr(infoMessage.c_str()));
+        AZ_TracePrintf("RobotImporterWidget", infoMessage.c_str());
```
lemonade-dm: This opens up the possibility of [format string injection](https://owasp.org/www-community/attacks/Format_string_attack) by directly using the input string as the format string.

A simpler method that isn't vulnerable to format string injection is to directly call the `AZ::Debug::Trace::Output` function which doesn't use a format string.
i.e
```c++
AZ::Debug::Trace::Instance().Output("RobotImporterWidget", infoMessage.c_str());
```

adamdbrw:
Good points suggestions of improvements, thank you! @michalpelka will submit fixes (he has taken the Importer under his wings), with my assistance where necessary.

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
michalpelka: Thanks for that hint!

Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
michalpelka: Fixing that needed to implement move constructor to CollidersMaker. Please take a look at  [that](https://github.com/RobotecAI/o3de-ros2-gem/commit/aefb1e49b84b5420155e5fac47315bd446505430#diff-1d0a02d3c5cc965628b4c5487fae26cf3de67ad9d6025b0631336ca7869a2383R37)

michalpelka: cc @adamdbrw

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
michalpelka: I keep unresolved, I activated RTTI and added assertions. I think it is not need to operate, but the bug in URDF parser can cause undefined behavior here without RTTI. Please take a look @adamdbrw

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
michalpelka: Thanks for that tip!

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
michalpelka: @adamdbrw please verify if this logic is what you designed.

michalpelka:
Thank you @lumberyard-employee-dm and @spham-amzn  for your insightful review.
I've added some changes that I feel need to be reviewed: move constructor `CollidersMaker::CollidersMaker(CollidersMaker&& other)` and modification of `VisualsMaker::VisualsMaker` that converts std container (from URDF lib) type to AZStd.
I also decided to activate RTTI, but it is to discuss.
@spham-amzn there is a lot of TODOs both from the code side, UX side - that are tracked in issues that I am adding lately.

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
spham-amzn: The use of RTTI is fine as long as this code is not client/game code.

Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
spham-amzn: Or another option is to make m_collidersMaker just a member
```
CollidersMaker m_collidersMaker;
```
and then it could be initialized like
```
        , m_visualsMaker(modelFilePath, model->materials_)
        , m_collidersMaker(modelFilePath)
```

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
adamdbrw: This is URDF parser (urdfdom package) specifics - there is a single element as well as an array for the same thing. Thus, the comment is correct

spham-amzn (approved):

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
lemonade-dm: Actually we don't disable disable RTTI explicitly anymore.
We use the default value for the compiler(which is on by default for most compilers).

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,218 @@
...
+    VisualsMaker::VisualsMaker(const AZ::IO::Path& modelPath, const std::map<std::string, urdf::MaterialSharedPtr>& materials)
+        : m_modelPath(modelPath)
+    {
+        std::for_each(
+            std::begin(materials),
+            std::end(materials),
+            [&](const auto& p)
+            {
+                m_materials[AZStd::string(p.first.c_str())] = p.second;
+            });
```
lemonade-dm: Nitpick: This could actually use AZStd::ranges::for_each function to reduce the need for using iteators.
Also it is slightly more efficient to also pass in the `p.first.size()` to the AZStd::string_constructor
```suggestion
        AZtd::ranges::for_each(
            materials,
            [&](const auto& p)
            {
                m_materials[AZStd::string(p.first.c_str(), p.first.size())] = p.second;
            });
```

lemonade-dm (approved):

PR #122, opened 2022-07-26
Get asset path from model path for mesh loader

pijaro:


Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
```diff
@@ -272,4 +269,38 @@ namespace ROS2
...
+        };
+
+        auto azmodelPath = AZStd::string(modelPath.replace_extension("azmodel").string().c_str());
+        AZStd::to_lower(azmodelPath.begin(), azmodelPath.end());
+
+        size_t watchedDirPose = AZStd::string::npos;
+        for (auto watchedDirectory : watchedDirectoriesNames)
+        {
+            AZStd::to_lower(watchedDirectory.begin(), watchedDirectory.end());
+            auto dirToWatch = AZStd::string("/") + watchedDirectory + AZStd::string("/");
```
adamdbrw: Better to use API for file paths (e.g. filesystem) for portability.

Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
```diff
@@ -272,4 +269,38 @@ namespace ROS2
...
+            watchedDirPose = azmodelPath.find(dirToWatch);
```
adamdbrw: Is this fully deterministic (e.g. directory name can be present in the root), e.g.:
modelPath: "/home/user/some_folder/projects_folder/project_name/Assets/model_folder/model.azmodel"
watchedDirectory1: "model_folder" (assumed in Assets?)
watchedDirectory2: "some_folder"
Both would match this find

Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
pijaro: No longer applicable - I changed the way how we look for an asset.

Code/Source/URDF/RobotImporter/URDFPrefabMaker.cpp
pijaro: No longer applicable - I changed the way how we look for an asset.

adamdbrw (approved):

PR #124, opened 2022-07-26
URDF importer: importer and it's GUI separation

piotr-zyskowski-rai:


Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
-    void RobotImporterWidget::OnModelLoaded()
+    void RobotImporterWidget::ReportError(AZStd::string errorMessage)
     {
-        m_robotNameLabel.setText(m_urdfModel->getName().c_str());
+        QMessageBox::critical(this, QObject::tr("Error"), errorMessage.c_str());
+        m_logLabel.setText(errorMessage.c_str());
+        AZ_Error("RobotImporterWidget", false, errorMessage.c_str());
+    }
+    void RobotImporterWidget::ReportWarning(AZStd::string warningMessage)
+    {
```
adamdbrw: No text is set for log label here. Is this intentional?
It would be best to extract this behavior to a dedicated utility class. Is there anything similar in O3DE already?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
-        m_importFileDialog.setNameFilter(QObject::tr("Unified Robot Description Format (*.urdf)"));
-        m_importFileDialog.setViewMode(QFileDialog::Detail);
+        mainLayout->addWidget(&m_logLabel);
+        mainLayout->addStretch();
 
         QObject::connect(
             selectFileButton,
             &QPushButton::clicked,
             this,
             [this]()
```
adamdbrw, line 49: no need to capture "this" anymore, only m_robotImporter is necessary

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
+        case QMessageBox::Cancel:
+            return ExistingPrefabAction::Cancel;
+        default:
+            return ExistingPrefabAction::Cancel;
+        }
+    }
+
+    AZStd::string RobotImporterWidget::GetPathWithExtension(AZStd::string extensionDescription, QFileDialog::FileMode mode)
+    {
+        QFileDialog m_importFileDialog;
```
adamdbrw: misleading name (m_ should be reserved for members)

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
+        m_importFileDialog.setFileMode(mode);
+        m_importFileDialog.setNameFilter(QObject::tr(extensionDescription.c_str()));
+        m_importFileDialog.setViewMode(QFileDialog::Detail);
+
+        int result = m_importFileDialog.exec();
+        if (result != QDialog::DialogCode::Accepted)
+        {
+            return {};
```
adamdbrw: Use AZStd::optional if failure is an option

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
+        msgBox.setButtonText(QMessageBox::Discard, "Save As..");
+        auto ret = msgBox.exec();
+        switch (ret)
+        {
+        case QMessageBox::Save:
+            return ExistingPrefabAction::Overwrite;
+        case QMessageBox::Discard:
+            return ExistingPrefabAction::NewName;
+        case QMessageBox::Cancel:
+            return ExistingPrefabAction::Cancel;
```
adamdbrw: [[fallthrough]];

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
+        default:
+            return ExistingPrefabAction::Cancel;
+        }
+    }
+
+    AZStd::string RobotImporterWidget::GetPathWithExtension(AZStd::string extensionDescription, QFileDialog::FileMode mode)
+    {
```
adamdbrw: All these GetPath functions are internal utilities - no need for them to be members. Consider putting them in Internal namespace or out in utils (if anything else could ever use them)

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
-            AZ_Error("RobotImporterWidget", false, "Importing robot definition failed with error: %s", outcome.GetError().c_str());
+    AZStd::string RobotImporterWidget::GetNewPrefabPath()
+    {
+        return GetPathWithExtension("Prefab (*.prefab)", QFileDialog::AnyFile);
+    }
+
+    RobotImporterInputInterface::ExistingPrefabAction RobotImporterWidget::GetExistingPrefabAction()
+    {
+        QMessageBox msgBox;
+        msgBox.setText("The prefab file already exists");
```
adamdbrw: UX texts - what would be default behavior for QObject::tr if we don't specify translation for this string precisely?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -27,45 +28,82 @@ namespace ROS2
...
+        msgBox.setInformativeText("Do you want to overwrite it or choose a new filename?");
+        msgBox.setStandardButtons(QMessageBox::Save | QMessageBox::Discard | QMessageBox::Cancel);
+        msgBox.setDefaultButton(QMessageBox::Save);
+        msgBox.setButtonText(QMessageBox::Save, "Overwrite");
```
adamdbrw: might be in QObject::tr dictionary

Code/Source/RobotImporter/URDF/RobotImporterInputInterface.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace ROS2
+{
+    class RobotImporterInputInterface
+    {
+    public:
+        enum ExistingPrefabAction
+        {
+            Overwrite,
+            NewName,
```
adamdbrw: CreateWithNewName

Code/Source/RobotImporter/URDF/RobotImporterInputInterface.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+#pragma once
+
+#include <AzCore/std/string/string.h>
+
+namespace ROS2
+{
+    class RobotImporterInputInterface
```
adamdbrw: By name alone, I would expect RobotImporter to derive from this interface. 
This interface seems to be a set of utils actually.
Is there any other sensible way of implementing this interface than what is done by the Widget? If no, perhaps these should be utilities / internal instead.

In the interface, is there any context required? For example, a parent QObject for a message box.

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Warning reporting was created to make a more "complete" reporting API (error, warn, info). Since I did not use it anywhere I got rid of it entirely. When it will be needed it can be brought back. Some more sophisticated warning reporting should be placed in GUI in the future.

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Changed to `&robotImporter = m_robotImporter`

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Removed the `Cancel` step entirely. Less code ;)

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Done :+1: (I mean I added the `QObject::tr` if that's what you meant)

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: I changed it to  `QObject::tr("Overwrite")`. 
I am not sure what you mean by "might be in QObject::tr dictionary". Could you point me to some resources?

Code/Source/RobotImporter/URDF/RobotImporterInputInterface.h
piotr-zyskowski-rai: Done :+1:

Code/Source/RobotImporter/URDF/RobotImporterInputInterface.h
piotr-zyskowski-rai: I changed the name to `RobotImporterUserInteractions` first but in the end, I decided to remove this entirely and simply pass the widget for whatever needs the user interactions (RobotImporter and URDFPrefabMaker at the moment)

> Is there any other sensible way of implementing this interface than what is done by the Widget? 

It could be implemented in a test environment as a user mockup but I am not sure whether such tests ever emerge. When it would be useful, this interface might be brought back

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: removed `m_` :+1:

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Moved them to internal space :+1:

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Applied this. Now the nullopt is returned when the user cancels. Also, this change led to a bit more changes, for example, I moved the path existence validation to the GUI side. IMO, it would be advisable to take a look at the entire code, again since it was moved around a bit.

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -6,18 +6,63 @@
...
 #include "RobotImporter/RobotImporterWidget.h"
-#include "RobotImporter/URDF/URDFPrefabMaker.h"
+
+#include <AzCore/Utils/Utils.h>
 
 namespace ROS2
 {
+    namespace Internal
+    {
+        AZStd::optional<QString> GetPathWithExtension(
```
adamdbrw: The name could be changed to something that better matches the behavior 
(interaction with a user through file dialog)

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -6,18 +6,63 @@
...
+            }
+
+            return importFileDialog.selectedFiles().first();
+        }
+
+        ExistingPrefabAction GetExistingPrefabAction(QWidget* parent = nullptr)
+        {
+            QMessageBox msgBox(parent);
+            msgBox.setWindowTitle(QObject::tr("Prefab file exists"));
+            msgBox.setText(QObject::tr("The prefab file already exists."));
```
adamdbrw: suggestion: add <name> to debug - so the user knows which prefab file, unless it is obvious in the context

Code/Source/RobotImporter/RobotImporterWidget.h
```diff
@@ -18,18 +18,37 @@
...
+    };
+
     //! Handles UI for the process of URDF importing
     class RobotImporterWidget : public QWidget
     {
         Q_OBJECT
     public:
         explicit RobotImporterWidget(QWidget* parent = nullptr);
 
+        //! Report an error to the user.
```
adamdbrw, line 29: Expand: "This will populate the log and show a window  (..)"

Code/Source/RobotImporter/URDF/RobotImporter.h
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <memory>
```
adamdbrw: not needed here apparently

Code/Source/RobotImporter/URDF/RobotImporter.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace ROS2
+{
+    class RobotImporterWidget;
+
+    class RobotImporter
+    {
+    public:
+        explicit RobotImporter(RobotImporterWidget& robotImporterWidget);
```
adamdbrw: RobotImporter could be designed so that it also work without the widget - for example, within the Asset Processor. 
This file could include RobotImporterConfiguration which would contain all that is necessary, except for the GUI element for logging. Consider using optional std::function(s) for logging, which could be set accordingly when constructing with the Widget (by default they would only log with to console with AZ_TracePrintf or similar)

adamdbrw: Ok looking further I see that there is still URDFPrefabMaker which could be used independently from the Widget. Consider points in the comment above but feel free to disregard.

adamdbrw (commented):
Big improvement - only a few comments remain.

Code/Source/RobotImporter/RobotImporterWidget.cpp
adamdbrw: https://doc.qt.io/qt-6/internationalization.html 
One needs to provide translations, but it is not necessary at this point.

PR #129, opened 2022-08-03
Urdf composition

adamdbrw:


Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -0,0 +1,174 @@
...
+                // Get asset path for a given model path
+                auto assetPath = PrefabMakerUtils::GetAssetPathFromModelPath(modelPath);
+
+                entity->CreateComponent(AZ::Render::EditorMeshComponentTypeId);
+                entity->Activate();
+                AZ::Render::MeshComponentRequestBus::Event(
+                    entityId, &AZ::Render::MeshComponentRequestBus::Events::SetModelAssetPath, assetPath.c_str());
+                entity->Deactivate();
+
+                // TODO apply scale
```
pijaro, line 131: Nitpick: I believe this TODO no longer applies.

pijaro (approved):

Code/Source/RobotImporter/URDF/VisualsMaker.cpp
adamdbrw: It does - it is a different scale, specified within urdf (not the mesh). The mesh resource can be rescaled. Example:
` <mesh filename="package://turtlebot4_description/meshes/rplidar.dae" scale="1 1 1"/>`

PR #139, opened 2022-08-10
A bit of refactor plus robot control.

adamdbrw:
This PR is a bit mixed, as it also contains some fixes / refactoring items. But it will all be subjected to proper feature review through the base branch.

### DIFF
--- a/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
+++ b/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
@@ -18,27 +18,31 @@
 namespace ROS2
 {
     void CollidersMaker::AddColliders(urdf::LinkSharedPtr link, AZ::EntityId entityId)
-    {
-        if (link->collision_array.size() == 0)
-        { // one or zero colliders - element is used
-            AddCollider(link->collision, entityId);
-            return;
+    { // TODO - refactor out naming (colliders, visuals)
+        auto colliderName = AZStd::string::format("%s_collider", link->name.c_str());
+        int nameSuffixIndex = 1; // For disambiguation when multiple unnamed colliders are present. The order does not matter here
+        for (auto collider : link->collision_array)
+        { // one or more colliders - the array is used
+            auto generatedName = link->visual_array.size() > 1
+                ? AZStd::string::format("%s_%d", colliderName.c_str(), nameSuffixIndex)
+                : colliderName;
+            nameSuffixIndex++;
+            AddCollider(collider, entityId, generatedName);
         }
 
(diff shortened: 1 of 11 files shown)


Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -18,34 +18,48 @@
...
 {
     void CollidersMaker::AddColliders(urdf::LinkSharedPtr link, AZ::EntityId entityId)
-    {
-        if (link->collision_array.size() == 0)
-        { // one or zero colliders - element is used
-            AddCollider(link->collision, entityId);
-            return;
+    { // TODO - refactor out naming (colliders, visuals)
+        auto colliderName = AZStd::string::format("%s_collider", link->name.c_str());
+        int nameSuffixIndex = 1; // For disambiguation when multiple unnamed colliders are present. The order does not matter here
```
pijaro, line 23: Nitpick: if it is an index, it would be better to start counting from `0`.

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
```diff
@@ -95,4 +97,27 @@ namespace ROS2
         AzToolsFramework::EditorEntityContextRequestBus::Broadcast(
             &AzToolsFramework::EditorEntityContextRequests::AddRequiredComponents, *entity);
     }
+
+    bool PrefabMakerUtils::HasCollider(AZ::EntityId entityId)
+    {
+        AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
+        return entity->FindComponent<PhysX::EditorColliderComponent>() != nullptr ||
+            entity->FindComponent<PhysX::EditorShapeColliderComponent>() != nullptr;
```
pijaro, line 105: Nitpick: Technically, there is also a third type of editor collider component, `PhysX::EditorHeightfieldColliderComponent`.

pijaro (approved):

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
adamdbrw: This is a displayed number, so indexing for human-readable names should start with 1
(we say - the first item, the second item)

Code/Source/RobotImporter/URDF/PrefabMakerUtils.cpp
adamdbrw: Yeah, but in fact after discussion with the physics team we are actually going to require specifically ECC (nothing else works for us)

PR #142, opened 2022-08-11
Mesh colliders

pijaro:
- Builds pxmeshes with `.assetinfo` files
- then creates prefab

### DIFF
--- a/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -9,8 +9,8 @@
 #include <AzCore/Utils/Utils.h>
 #include <QMessageBox>
 #include <QPushButton>
-#include <QVBoxLayout>
 #include <QTimer>
+#include <QVBoxLayout>
 
 #include "RobotImporter/RobotImporterWidget.h"
 #include "RobotImporter/RobotImporterWidgetUtils.h"
@@ -31,16 +31,17 @@ namespace ROS2
     RobotImporterWidget::RobotImporterWidget(QWidget* parent)
         : QWidget(parent)
         , m_statusLabel("", this)
-        , m_robotImporter(
-            [this](const AZStd::string& message)
-            {
-              ReportInfo(message);
-            },
-            [this](const AZStd::string& message)
-            {
-              ReportError(message);
-            })
+        , m_selectFileButton(new QPushButton(this))
         , m_importerUpdateTimer(new QTimer(this))
+        , m_robotImporter(
+              [this](const AZStd::string& message)
+              {
+                  ReportInfo(message);
+              },
+              [this](const AZStd::string& message)
+              {
+                  ReportError(message);
+              })
     {
         setWindowTitle(QObject::tr("Robot definition file importer"));
         QVBoxLayout* mainLayout = new QVBoxLayout(this);
@@ -48,8 +49,7 @@ namespace ROS2
         QLabel* captionLabel = new QLabel(QObject::tr("Select a robot definition (URDF) file to import"), this);
         captionLabel->setWordWrap(true);
         mainLayout->addWidget(captionLabel);
-        QPushButton* selectFileButton = new QPushButton(QObject::tr("Load"), this);
-        mainLayout->addWidget(selectFileButton);
+        mainLayout->addWidget(m_selectFileButton);
         mainLayout->addWidget(&m_statusLabel);
         mainLayout->addStretch();
 
@@ -76,26 +76,28 @@ namespace ROS2
                     ReportError("User cancelled");
                     return;
                 }
-                m_robotImporter.Import({ urdfPath.value(), prefabPath.value()});
+                m_robotImporter.Import({ urdfPath.value(), prefabPath.value() });
                 m_selectFileButton->setEnabled(false);
                 m_importerUpdateTimer->start(500);
             });
         setLayout(mainLayout);
-
     }
 
-    void RobotImporterWidget::ImporterTimerUpdate() {
-        m_robotImporter.Update([this]() {
-                                    m_importerUpdateTimer->stop();
-                                    m_selectFileButton->setEnabled(true);
-                               });
+    void RobotImporterWidget::ImporterTimerUpdate()
+    {
+        m_robotImporter.Update(
+            [this]()
+            {
+                m_importerUpdateTimer->stop();
+                m_selectFileButton->setEnabled(true);
+            });
     }
 
     void RobotImporterWidget::ReportError(const AZStd::string& errorMessage)
     {
-         QMessageBox::critical(this, QObject::tr("Error"), QObject::tr(errorMessage.c_str()));
-         m_statusLabel.setText(errorMessage.c_str());
-         AZ_Error("RobotImporterWidget", false, errorMessage.c_str());
+        QMessageBox::critical(this, QObject::tr("Error"), QObject::tr(errorMessage.c_str()));
+        m_statusLabel.setText(errorMessage.c_str());
+        AZ_Error("RobotImporterWidget", false, errorMessage.c_str());
     }
 
     void RobotImporterWidget::ReportInfo(const AZStd::string& infoMessage)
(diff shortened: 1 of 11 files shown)


Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -63,20 +76,23 @@ namespace ROS2
...
-                    {
-                        ReportInfo(message);
-                    },
-                    [this](const AZStd::string& message)
-                    {
-                        ReportError(message);
-                    });
+                m_robotImporter.Import({ urdfPath.value(), prefabPath.value() });
+                m_selectFileButton->setEnabled(false);
+                m_importerUpdateTimer->start(500);
```
adamdbrw: Comment on this please - are we prohibiting user from running this twice by mistake, or something else? Why 500 - and perhaps using a named constant or comment to make clear it is miliseconds?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -63,20 +76,23 @@ namespace ROS2
...
             });
         setLayout(mainLayout);
     }
 
+    void RobotImporterWidget::ImporterTimerUpdate()
+    {
+        m_robotImporter.Update(
+            [this]()
+            {
+                m_importerUpdateTimer->stop();
```
adamdbrw: we could instead set this timer to single shot

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -30,20 +31,32 @@ namespace ROS2
...
         QLabel* captionLabel = new QLabel(QObject::tr("Select a robot definition (URDF) file to import"), this);
         captionLabel->setWordWrap(true);
         mainLayout->addWidget(captionLabel);
-        QPushButton* selectFileButton = new QPushButton(QObject::tr("Load"), this);
-        mainLayout->addWidget(selectFileButton);
+        mainLayout->addWidget(m_selectFileButton);
         mainLayout->addWidget(&m_statusLabel);
         mainLayout->addStretch();
 
+        connect(m_importerUpdateTimer, &QTimer::timeout, this, QOverload<>::of(&RobotImporterWidget::ImporterTimerUpdate));
```
adamdbrw: Could we simply use a lambda here and do away with the member function? The name indicates it is only called from a timer.

Code/Source/RobotImporter/RobotImporterWidget.h
```diff
@@ -36,7 +37,12 @@ namespace ROS2
         //! @param infoMessage info message to display to the user
         void ReportInfo(const AZStd::string& infoMessage);
 
-    private:
         QLabel m_statusLabel;
+        QPushButton* m_selectFileButton;
```
adamdbrw: could these be on stack? What are the considerations here?

Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
```diff
@@ -21,18 +21,34 @@ namespace ROS2
...
         VisualsMaker m_visualsMaker;
-        CollidersMaker m_collidersMaker;
+        AZStd::shared_ptr<CollidersMaker> m_collidersMaker;
         InertialsMaker m_inertialsMaker;
         JointsMaker m_jointsMaker;
+
+        BuildReadyCallback m_notifyBuildReadyCb;
+
+        AZStd::atomic_bool m_stopBuildFlag;
+        AZStd::thread m_buildThread;
```
adamdbrw, line 52: All the "build" things could be moved out of the class to a separate one, to encapsulate this behavior. Make an issue for this.

adamdbrw: likely we could do away with m_colliders being a shared_ptr if we do that.

Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -25,12 +27,73 @@ namespace ROS2
...
+                                if (assetId.IsValid())
+                                {
+                                    AZ_Printf("CollisionMaker", "Asset %s found and valid...", iter->c_str());
+                                    m_collidersMaker->m_meshesToBuild.erase(iter--);
+                                }
+                            }
+                        }
+                        if (!m_collidersMaker->m_meshesToBuild.empty())
+                        {
+                            AZStd::this_thread::sleep_for(AZStd::chrono::milliseconds(1000));
```
adamdbrw: Too much time to wait - 50 ms is enough since we are only checking a stop flag. This could also be resolved with a conditional.

Code/Source/RobotImporter/URDF/RobotImporter.h
```diff
@@ -19,12 +22,19 @@ namespace ROS2
...
-    namespace RobotImporter
+    class RobotImporter
     {
-        void Import(
-            const RobotImporterConfig& config,
-            std::function<void(const AZStd::string&)> infoLogger,
-            std::function<void(const AZStd::string&)> errorLogger);
-    }
+    public:
+        RobotImporter(std::function<void(const AZStd::string&)> infoLogger,
```
adamdbrw: one function with 2 params (log, level) is typically used.

Code/Source/RobotImporter/URDF/RobotImporter.h
```diff
@@ -19,12 +22,19 @@ namespace ROS2
...
+                      std::function<void(const AZStd::string&)> errorLogger);
+        void Import(const RobotImporterConfig& config);
```
adamdbrw: Document - what does this function do?

Code/Source/RobotImporter/URDF/RobotImporter.h
```diff
@@ -19,12 +22,19 @@ namespace ROS2
...
+        void Update(std::function<void()> importFinishedCb);
```
adamdbrw: As above - this is unclear, especially why a function is used here.

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -63,20 +76,23 @@ namespace ROS2
...
         setLayout(mainLayout);
     }
 
+    void RobotImporterWidget::ImporterTimerUpdate()
+    {
+        m_robotImporter.Update(
+            [this]()
+            {
+                m_importerUpdateTimer->stop();
+                m_selectFileButton->setEnabled(true);
```
adamdbrw: Could we simply set this button status more directly on timer elapse?

adamdbrw (commented):
Some changes to apply immediately, some can be delegated to further issues

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Not really. There is no time that we could set here to ensure that the import is done that will not be ridiculously high. I believe a regular timer that polls in a set interval is the only option. The interval time is up to discussion, however.

Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
piotr-zyskowski-rai: Done https://github.com/RobotecAI/o3de-ros2-gem/issues/151

Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
piotr-zyskowski-rai: I changed the time to 50. The issue that will trace the threading improvement is #151

Code/Source/RobotImporter/RobotImporterWidget.cpp
piotr-zyskowski-rai: Not really: https://github.com/RobotecAI/o3de-ros2-gem/pull/142#discussion_r954939608

Code/Source/RobotImporter/URDF/RobotImporter.h
piotr-zyskowski-rai: Changed the name to be more informative and added docs

Code/Source/RobotImporter/URDF/RobotImporter.h
piotr-zyskowski-rai: Changed the name to be more informative and added docs

adamdbrw (approved):

PR #149, opened 2022-08-23
Set up namespaces before setting up transform publishing

### LINKED ISSUE #148: Incorrect publishing of dynamic transforms of namespaced ROS2 Frames
The `ROS2 Frame` of dynamic entities with namespaced frame name are not published correctly for non-topLevel entities. 

## Expected Behavior
Dynamic transform between root frame `odom` and frame `base_link2` in Default namespace for non-topLevel entity with name `Robot` should be set up as `geometry_msgs::msg::TransformStamped` with 
```
t.header.frame_id = "odom";
t.child_frame_id = "Robot//base_link2";
```

## Actual Behavior
Dynamic transform between root frame `odom` and frame `base_link2` in Default namespace for non-topLevel entity with name `Robot` is set up as `geometry_msgs::msg::TransformStamped` with 
```
t.header.frame_id = "odom";
t.child_frame_id = "base_link2";
```

## Example
image

lgleim:
Fixes https://github.com/RobotecAI/o3de-ros2-gem/issues/148

image

### DIFF
--- a/Code/Source/Frame/ROS2FrameComponent.cpp
+++ b/Code/Source/Frame/ROS2FrameComponent.cpp
@@ -18,8 +18,11 @@ namespace ROS2
 {
     void ROS2FrameComponent::Activate()
     {
+        m_namespaceConfiguration.PopulateNamespace(IsTopLevel(), GetEntity()->GetName());
+        
         if (m_publishTransform)
         {
+            AZ_TracePrintf("ROS2FrameComponent", "Setting up %s transfrom between parent %s and child %s to be published %s", IsDynamic() ? "dynamic" : "static", GetParentFrameID().data(), GetFrameID().data(), IsDynamic() ? "continuously to /tf" : "once to /tf_static");
             m_ros2Transform = AZStd::make_unique<ROS2Transform>(GetParentFrameID(), GetFrameID(), IsDynamic());
             if (IsDynamic())
             {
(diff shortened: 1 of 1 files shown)


Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -18,8 +18,11 @@ namespace ROS2
 {
     void ROS2FrameComponent::Activate()
     {
+        m_namespaceConfiguration.PopulateNamespace(IsTopLevel(), GetEntity()->GetName());
+        
         if (m_publishTransform)
         {
+            AZ_TracePrintf("ROS2FrameComponent", "Setting up %s transfrom between parent %s and child %s to be published %s", IsDynamic() ? "dynamic" : "static", GetParentFrameID().data(), GetFrameID().data(), IsDynamic() ? "continuously to /tf" : "once to /tf_static");
```
adamdbrw: There is a .clang_format file in the project, which can be used to check formatting. This line should be wrapped otherwise it will fail our auto check.

adamdbrw (changes requested):
Looks good, thank you for fixing this!

Only one minor issue with formatting so that our checks can pass.

adamdbrw (approved):

lgleim:
Apparently empty lines with spaces are handled inconsistently across clang-format versions. Anyway, third time is the charm @adamdbrw

Issue #33, opened 2022-05-03
Camera sensor PoC


adamdbrw:
@piotr-zyskowski-rai please design and fill in the requirements. A starting point:
- [x] Create a camera component
- [x] Acquire frames from a given transform at a determined rate
- [x] Support multiple cameras in the scene
- [x] Investigate performance
- [x] Publish raw image, visible in RViz
- [x] Publish camera info message 
- [x] Enable setting of FoV
- [x] Enable setting of resolution

adamdbrw:
@piotr-zyskowski-rai is this one done? What is the status?

piotr-zyskowski-rai:
I have created PR #154 to complete the "Publish camera info message" part of this task

piotr-zyskowski-rai:
PR #154 was merged so this task can be closed

PR #154, opened 2022-09-01
Camera info topic published (#33)

piotr-zyskowski-rai:
Implemented camera info publishing. 

Because of current camera sensor features, the provided information is limited to:
- image width and height
- camera intrinsics
- distortion type

### DIFF
--- a/Code/Source/Camera/CameraSensor.cpp
+++ b/Code/Source/Camera/CameraSensor.cpp
@@ -24,13 +24,13 @@
 namespace ROS2
 {
     CameraSensorDescription::CameraSensorDescription(const AZStd::string& cameraName, float verticalFov, int width, int height)
-        : verticalFieldOfViewDeg(verticalFov)
-        , width(width)
-        , height(height)
-        , cameraName(cameraName)
-        , aspectRatio(static_cast<float>(width) / static_cast<float>(height))
-        , viewToClipMatrix(MakeViewToClipMatrix())
-        , cameraIntrinsics(MakeCameraIntrinsics())
+        : m_verticalFieldOfViewDeg(verticalFov)
+        , m_width(width)
+        , m_height(height)
+        , m_cameraName(cameraName)
+        , m_aspectRatio(static_cast<float>(width) / static_cast<float>(height))
+        , m_viewToClipMatrix(MakeViewToClipMatrix())
+        , m_cameraIntrinsics(MakeCameraIntrinsics())
     {
         validateParameters();
     }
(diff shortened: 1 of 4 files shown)


Code/Source/Camera/CameraSensor.cpp
```diff
@@ -30,6 +30,7 @@ namespace ROS2
         , cameraName(cameraName)
```
adamdbrw: with all class / struct member fields, follow m_<name> convention. It makes it easier to determine what is a part of state and what is an argument / local variable

Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -45,6 +46,7 @@ namespace ROS2
         void FrequencyTick() override;
 
         std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::Image>> m_imagePublisher;
+        std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::CameraInfo>> m_cameraInfoPublisher;
```
adamdbrw, line 49: Could it be an unique_ptr instead? (Same above).

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -49,7 +50,18 @@ namespace ROS2
         AZ_Assert(!cameraName.empty(), "Camera name cannot be empty");
     }
 
+    std::array<double, 9> CameraSensorDescription::MakeCameraIntrinsics() const
+    {
```
adamdbrw, line 56: A comment on what is happening here would be useful

adamdbrw (commented):
Few minor comments, take a look

Code/Source/Camera/ROS2CameraSensorComponent.h
piotr-zyskowski-rai: Since this is a publisher, it is created by rclcpp node API
```
        auto ros2Node = ROS2Interface::Get()->GetNode();
        ...
        m_imagePublisher =
            ros2Node->create_publisher<sensor_msgs::msg::Image>(cameraImageFullTopic.data(), cameraImagePublisherConfig.GetQoS());
```
And it returns `shared_ptr` and there is no straightforward way to make `unique_ptr` from it

Code/Source/Camera/CameraSensor.cpp
piotr-zyskowski-rai: Added comment :+1: Is this level of detail ok?

adamdbrw: Perfect, thank you

adamdbrw (approved):

PR #156, opened 2022-09-02
Product path query

piotr-zyskowski-rai:


michalpelka:
This PR should also cover this case:
https://github.com/RobotecAI/o3de-ros2-gem/issues/165

Code/Source/RobotImporter/URDF/UrdfParser.cpp
```diff
@@ -19,6 +19,7 @@ namespace ROS2
 {
     urdf::ModelInterfaceSharedPtr UrdfParser::Parse(const AZStd::string& xmlString)
     {
+        std::setlocale(LC_NUMERIC, "en_US.UTF-8");
```
adamdbrw: Unrelated change from another PR

Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -24,6 +25,86 @@
...
+        AZ::IO::Path GetFullURDFMeshPath(AZ::IO::Path modelPath, AZ::IO::Path meshPath)
+        {
+            modelPath.RemoveFilename();
+            AZ::StringFunc::Replace(meshPath.Native(), "package://", "", true, true);
+            modelPath /= meshPath;
+
+            return modelPath;
+        }
+
+        AZStd::optional<AZ::IO::Path> GetMeshProductPathFromSourcePath(const AZ::IO::Path& sourcePath)
```
adamdbrw, line 40: This looks like a Utils function - not related to Colliders. Best to separate it out.

PR #158, opened 2022-09-05
design docs and a diagram for vehicle dynamics

adamdbrw:
Fixes #144

Signed-off-by: Adam Dabrowski <adam.dabrowski@robotec.ai>

PR #162, opened 2022-09-07
Vehicle dynamics

adamdbrw:
Vehicle Dynamics, following the [design document](https://github.com/RobotecAI/o3de-ros2-gem/pull/158).

Tested and ready.


adamdbrw:
@michalpelka please review logic (algorithms, correctness) only for now. Primary focus is the SimpleDriveModel

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,98 @@
...
+                desiredAngularSpeedX,
+                currentAngularSpeedX,
+                pidCommand);
+
+            if (AZ::IsClose(pidCommand, 0.0)) // TODO - use the third argument with some reasonable value which means "close enough"
+            {
+                continue;
+            }
+
+            auto impulse = pidCommand;
```
michalpelka: We should limit the torque here

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+            auto transformedTorqueVector = wheelTransform.TransformVector(AZ::Vector3(impulse, 0, 0));
```
michalpelka: I do not like this direction axis. In design correct axis is -Z for left and +Z for right wheel IHMO

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,98 @@
...
+            Physics::RigidBodyRequestBus::Event(wheelEntity, &Physics::RigidBodyRequests::ApplyAngularImpulse, transformedTorqueVector);
```
michalpelka, line 140: Here is issue that also @pawelbudziszewski raised. We assume that we have constant frame rate. 
I can see that we apply elapsed time to PID controller - since it need that for I and D -action.
From this code it seems that we will apply less momentum when our simulation will run with 30 fps.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,98 @@
...
+        for (auto& wheelEntity : wheelEntities)
+        {
+            AZ::Transform wheelTransform;
+            AZ::TransformBus::EventResult(wheelTransform, wheelEntity, &AZ::TransformBus::Events::GetWorldTM);
+
+            AZ::Transform inverseWheelTransform = wheelTransform.GetInverse();
+            AZ::Vector3 currentAngularVelocity;
+            Physics::RigidBodyRequestBus::EventResult(currentAngularVelocity, wheelEntity, &Physics::RigidBodyRequests::GetAngularVelocity);
+            currentAngularVelocity = inverseWheelTransform.TransformVector(currentAngularVelocity);
+            auto currentAngularSpeedX = currentAngularVelocity.GetX();
```
michalpelka: I think it should be X. I think that we should add constexpr on axis of wheel to measure velocity and apply it correctly.

michalpelka (commented):
More or less high level ideas and questions:
 - we probably need to add ramp generator for control values
 - why we need to apply torque everywhere e (consider applying velocity)
 - There is some issue with rotation speed computation.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
pawelbudziszewski: In the PoC I solved this by multiplying the impulse by deltaTime. It would be something like this:
`auto impulse = pidCommand * nsDt;`

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
pawelbudziszewski: Especially concerning the fact, that each wheel may be rotated differently (which is the case in the physics test scene)

michalpelka: In [ee4620e](https://github.com/RobotecAI/o3de-ros2-gem/pull/162/commits/ee4620e43bd72ee7184540f630e96374ccfc1d9d) I added temporarily the direction.
In other words it is not hardcoded anymore. The ideal solution would be to find this axis from coordinate frames of joints.
If we reach the ideal solution, `m_drive_dir` and `m_steering_dir` won't be exposed to the user.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
michalpelka: Added as @pawelbudziszewski suggested.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
michalpelka: It will be added in the next iteration. Created #201

Code/Source/RobotControl/Controllers/AckermannController/AckermannControlComponent.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    void AckermannControlComponent::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<AckermannControlComponent, AZ::Component>()->Version(1);
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<AckermannControlComponent>("Ackermann Control", "Relays Ackermann commands to vehicle inputs")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
```
pijaro, line 25: Maybe add a category as well?
```cpp
->Attribute(AZ::Edit::Attributes::Category, "ROS2")
```

Code/Source/VehicleDynamics/AxleConfiguration.cpp
```diff
@@ -0,0 +1,75 @@
...
+                        "Is it a drive axle",
+                        "Is this axle used for drive (all attached wheels)")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &AxleConfiguration::m_wheelRadius,
+                        "Wheel radius",
+                        "Radius of each wheel attached to axle")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &AxleConfiguration::m_axleWheels,
```
pijaro, line 53: We should tell the user that we expect the left wheel on the 1st element and the right on the 2nd. (comment in the code is not sufficient for the editor user)

adamdbrw:
Quite an important PR - I have added multiple reviewers but please do not feel obliged.

Code/Source/VehicleDynamics/AxleConfiguration.h
```diff
@@ -0,0 +1,37 @@
...
+        //! Helper functions for wheel entities. If there is only one wheel, same value is returned from both.
+        //! If no wheels are set, this will return invalid EntityId (check with IsValid())
+        AZ::EntityId GetLeftWheelEntityId() const; //! Return left-most wheel of the axis.
+        AZ::EntityId GetRightWheelEntityId() const; //! Return right-most wheel of the axis.
+
+        AZStd::string m_axleTag; //! Useful to differentiate between axles, can be empty.
+        AZStd::vector<AZ::EntityId> m_axleWheels; //! One or more wheels attached to this axle (typically 2), sorted left to right.
+
+        float m_wheelRadius = 0.35f; // TODO - this could be deduced from a model for a better default
+        bool m_isSteering = false;
```
pijaro, line 34: nit: maybe `m_isSteerable` would be a bit more informative?

Code/Source/RobotControl/Controllers/RigidBodyController/RigidBodyTwistControlComponent.cpp
```diff
@@ -0,0 +1,68 @@
...
+{
+    void RigidBodyTwistControlComponent::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<RigidBodyTwistControlComponent, AZ::Component>()->Version(1);
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<RigidBodyTwistControlComponent>("Rigid Body Twist Control", "Simple control through RigidBody")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
```
pijaro, line 26: We can add a category here as well.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        const double nsDtSec = double(nsDt) / 1e9;
+        for (const auto& steeringElementData : m_steeringData)
+        {
+            auto steeringEntity = steeringElementData.m_steeringEntity;
+            AZ::Vector3 currentSteeringElementRotation;
+            AZ::TransformBus::EventResult(currentSteeringElementRotation, steeringEntity, &AZ::TransformBus::Events::GetLocalRotation);
+            auto currentSteeringAngle = currentSteeringElementRotation.Dot(steeringElementData.m_turnAxis);
+            double pidCommand = m_steeringPid.ComputeCommand(steering - currentSteeringAngle, nsDt);
+            AZ_TracePrintf(
```
pijaro: Isn't trace print a bit too intrusive in this case?

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,149 @@
...
+            {
+                const float defaultFloatRadius = 0.35f;
+                AZ_Warning("ApplySpeed", false, "Wheel radius is zero (or too close to zero), resetting to default %f", defaultFloatRadius);
+                wheelRadius = defaultFloatRadius;
+            }
+
+            auto desiredAngularSpeedX = speed / wheelRadius;
+
+            double pidCommand = m_speedPid.ComputeCommand(desiredAngularSpeedX - currentAngularSpeedX, nsDt);
+            AZ_TracePrintf(
```
pijaro: A bit of the trace print flood here as well :smiley:

Code/Source/VehicleDynamics/VehicleInputsState.h
```diff
@@ -0,0 +1,66 @@
...
+        int64_t m_timeoutUs;
+        int64_t m_lastUpdateUs = 0;
+        T m_input = 0;
+    };
+
+    //! The most recent inputs
+    struct VehicleInputsState
+    {
+        InputZeroedOnTimeout<float> m_speed; // Mps
+        InputZeroedOnTimeout<float> m_steering; // Radians, right is +, left is -
```
pijaro: According to ROS2 [REP-103](https://www.ros.org/reps/rep-0103.html) rotating right (clockwise) is `-`, whereas left (counter-clockwise) is `+` (right hand rule).

Code/Source/VehicleDynamics/VehicleModelComponent.cpp
```diff
@@ -0,0 +1,111 @@
...
+            serialize->Class<VehicleModelComponent, AZ::Component>()
+                ->Version(2)
+                ->Field("ChassisConfiguration", &VehicleModelComponent::m_chassisConfiguration)
+                ->Field("DriveModel", &VehicleModelComponent::m_driveModel)
+                ->Field("VehicleModelLimits", &VehicleModelComponent::m_vehicleLimits);
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<VehicleModelComponent>("Vehicle Model", "Customizable vehicle model component")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
```
pijaro, line 53: +Category ?

Code/Source/VehicleDynamics/VehicleModelLimits.h
```diff
@@ -0,0 +1,28 @@
...
+    {
+    public:
+        AZ_TYPE_INFO(VehicleModelLimits, "{76DA392D-64BB-45A8-BC90-84AAE7901991}");
+        VehicleModelLimits() = default;
+        static void Reflect(AZ::ReflectContext* context);
+
+        static float LimitValue(float value, float absoluteLimit);
+
+        float m_speedLimitMps = 10.0f; //! Applies to absolute value
+        float m_steeringLimitRads = 0.7f; //! Applies to absolute value
```
pijaro: Any particular reason why there are unit types in variables name?
```suggestion
        float m_speedLimit = 10.0f; //! [Mps] Applies to absolute value
        float m_steeringLimit = 0.7f; //! [rad] Applies to absolute value
```

Code/Source/VehicleDynamics/WheelControllerComponent.cpp
```diff
@@ -0,0 +1,80 @@
...
+                ->Version(1)
+                ->Field("SteeringEntity", &WheelControllerComponent::m_steeringEntity)
+                ->Field("VisualEntity", &WheelControllerComponent::m_visualEntity)
+                ->Field("DriveDir", &WheelControllerComponent::m_driveDir)
+                ->Field("SteeringDir", &WheelControllerComponent::m_steeringDir);
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<WheelControllerComponent>("Wheel controller", "Handle wheel physics")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
```
pijaro, line 54: +Category?

Code/Source/VehicleDynamics/VehicleModelLimits.h
```diff
@@ -0,0 +1,28 @@
...
+    struct VehicleModelLimits
+    {
+    public:
+        AZ_TYPE_INFO(VehicleModelLimits, "{76DA392D-64BB-45A8-BC90-84AAE7901991}");
+        VehicleModelLimits() = default;
+        static void Reflect(AZ::ReflectContext* context);
+
+        static float LimitValue(float value, float absoluteLimit);
+
+        float m_speedLimitMps = 10.0f; //! Applies to absolute value
```
pijaro, line 25: I see that everywhere `m/s` is a default, why `Mps` here?

Code/Source/RobotControl/Controllers/AckermannController/AckermannControlComponent.cpp
adamdbrw: This is a good idea - but also applies to all of our components. I created an [issue](https://github.com/RobotecAI/o3de-ros2-gem/issues/223) for this.

Code/Source/RobotControl/Controllers/RigidBodyController/RigidBodyTwistControlComponent.cpp
adamdbrw: https://github.com/RobotecAI/o3de-ros2-gem/issues/223 - will handle that for the Gem

Code/Source/VehicleDynamics/VehicleModelComponent.cpp
adamdbrw: https://github.com/RobotecAI/o3de-ros2-gem/issues/223

Code/Source/VehicleDynamics/AxleConfiguration.h
adamdbrw: "Is it a steering axle" shortened to isSteering.
Note that this is explained in the Editor context: "Is this axle used for steering (all attached wheels)"

Code/Source/RobotControl/Ackermann/AckermannBus.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+#pragma once
+
+#include "AckermannCommandStruct.h"
+#include <AzCore/EBus/EBus.h>
+#include <AzCore/RTTI/BehaviorContext.h>
+
+namespace ROS2
+{
+    //! Interface class for handling Ackermann commands through EBus notifications.
```
piotr-zyskowski-rai: `Ackerman commands` seem a bit non-technically-correct - it's a short name for `Ackerman kinematics steering commands`. 

Overall I would be in favor of changing Ackermann to AckermannKinematics, AckermannKinematicsSteering/Drive etc. in the code itself, but I understand that it is well accepted in the robotics community to call it in the shortened way. In the documentation, however, I would change this to its full name.

Code/Source/VehicleDynamics/Utilities.cpp
```diff
@@ -0,0 +1,125 @@
...
+    {
+        return Create2WheelAxle(leftWheel, rightWheel, "Rear", wheelRadius, false, true);
+    }
+
+    AZStd::vector<VehicleDynamics::SteeringDynamicsData> GetAllSteeringEntitiesData(const ChassisConfiguration& chassisConfig)
+    {
+        AZStd::vector<VehicleDynamics::SteeringDynamicsData> steeringEntitiesAndAxis;
+        for (const auto& axle : chassisConfig.m_axles)
+        {
+            if (axle.m_isSteering)
```
piotr-zyskowski-rai: early `continue` might be handy here

Code/Source/VehicleDynamics/Utilities.cpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return steeringEntitiesAndAxis;
+    }
+
+    AZStd::vector<VehicleDynamics::WheelDynamicsData> GetAllDriveWheelsData(const ChassisConfiguration& chassisConfig)
+    {
+        AZStd::vector<VehicleDynamics::WheelDynamicsData> driveWheelEntities;
+        for (const auto& axle : chassisConfig.m_axles)
+        {
+            if (axle.m_isDrive)
```
piotr-zyskowski-rai: Early continue might be handy here

Code/Source/VehicleDynamics/VehicleInputsState.h
```diff
@@ -0,0 +1,66 @@
...
+        int64_t m_timeoutUs;
+        int64_t m_lastUpdateUs = 0;
+        T m_input = 0;
+    };
+
+    //! The most recent inputs
+    struct VehicleInputsState
+    {
+        InputZeroedOnTimeout<float> m_speed; // Mps
+        InputZeroedOnTimeout<float> m_steering; // Radians, right is -, left is +
```
piotr-zyskowski-rai: Since it is documented anyway, use Doxygen member documentation syntax `//!<` and add a bit more details
Also, it would be more clear when `Mps` would be `m/s` :)

Code/Source/VehicleDynamics/VehicleModelComponent.cpp
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    void VehicleModelComponent::SetTargetSteeringFraction(float steeringFraction)
+    {
+        m_inputsState.m_steering.UpdateValue(steeringFraction * m_vehicleLimits.m_steeringLimit);
+    }
+
+    void VehicleModelComponent::OnTick(float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
+    {
+        uint64_t deltaTimeNs = deltaTime * 1000000000;
```
piotr-zyskowski-rai:
```suggestion
        uint64_t deltaTimeNs = deltaTime * 1'000'000'000;
```

Code/Source/VehicleDynamics/VehicleModelLimits.h
```diff
@@ -0,0 +1,28 @@
...
+    struct VehicleModelLimits
+    {
+    public:
+        AZ_TYPE_INFO(VehicleModelLimits, "{76DA392D-64BB-45A8-BC90-84AAE7901991}");
+        VehicleModelLimits() = default;
+        static void Reflect(AZ::ReflectContext* context);
+
+        static float LimitValue(float value, float absoluteLimit);
+
+        float m_speedLimit = 10.0f; //! [Mps] Applies to absolute value
```
piotr-zyskowski-rai: `m/s` is more clear IMHO. Also, for members, `//!<` is the correct syntax I think

Code/Source/VehicleDynamics/VehicleModelLimits.h
```diff
@@ -0,0 +1,28 @@
...
+{
+    //! A structure holding limits of vehicle, including speed and steering limits
+    struct VehicleModelLimits
+    {
+    public:
+        AZ_TYPE_INFO(VehicleModelLimits, "{76DA392D-64BB-45A8-BC90-84AAE7901991}");
+        VehicleModelLimits() = default;
+        static void Reflect(AZ::ReflectContext* context);
+
+        static float LimitValue(float value, float absoluteLimit);
```
piotr-zyskowski-rai, line 27: I would be nice to document this function

Code/Source/VehicleDynamics/WheelControllerComponent.h
```diff
@@ -0,0 +1,37 @@
...
+        static void GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required);
+        static void GetProvidedServices(AZ::ComponentDescriptor::DependencyArrayType& provided);
+        static void GetIncompatibleServices(AZ::ComponentDescriptor::DependencyArrayType& incompatible);
+
+        static void Reflect(AZ::ReflectContext* context);
+
+        AZ::EntityId m_steeringEntity; //! Rigid body to apply torque to. TODO - parent, this entity or custom.
+        AZ::EntityId m_visualEntity; //! Primary visual mesh or primitive for the wheel. It will be rotated according to the motion.
+        AZ::Vector3 m_driveDir{ 0.0, 0.0, 1.0 }; //! The direction of torque applied to wheel entity when speed is applied
+        AZ::Vector3 m_steeringDir{ 0.0, 0.0, 1.0 }; //! The direction of torque applied to steering entity when steering is applied
```
piotr-zyskowski-rai: `//!<` syntax for members

Code/ros2_files.cmake
```diff
@@ -4,69 +4,98 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
 set(FILES
-    Source/Camera/CameraSensor.cpp
```
piotr-zyskowski-rai, line 7: Nitpick: Wasn't the old indentation correct? Also, it would be easier to track the difference if we keep it

Code/Source/VehicleDynamics/WheelDynamicsData.h
```diff
@@ -0,0 +1,25 @@
...
+    {
+        AZ::EntityId m_wheelEntity;
+        AZ::Vector3 m_driveAxis;
+        float m_wheelRadius;
+    };
+
+    struct SteeringDynamicsData
+    {
+        AZ::EntityId m_steeringEntity;
+        AZ::Vector3 m_turnAxis;
```
piotr-zyskowski-rai: A bit of documentation might be in order

README.md
```diff
@@ -10,6 +10,13 @@ This Gem enables users to develop robotic simulations through ROS2 tools and com
   * [ROS 2 Galactic](https://docs.ros.org/en/galactic/Installation.html) with Ubuntu 20.04.
   * [ROS 2 Humble](https://docs.ros.org/en/humble/Installation.html) with Ubuntu 22.04.
 
+#### Additional ros packages required
+
+Replace `<distro>` with ROS 2 distribution name (galactic, humble, ..).
+
+* Ackermann messages: `sudo apt install ros-<distro>-ackermann-msgs`
+* Control toolbox `sudo apt install ros-<distro>-control-toolbox`
+
```
piotr-zyskowski-rai: Maybe it would be more convenient for the user to have two separate sections for each distribution?

Also, ROS2 Gem probably requires some more packages to work, which might not be available in ROS2 installation (like `urdfdom`, `sensor_msgs` etc), then maybe it would be useful to add it all here? Unless of course, they are in the ROS2 Desktop installation. In this case, however, it should be clearly mentioned that the installation should be a Desktop one.

````suggestion
#### Additional ROS 2 packages required

##### Galactic

```
sudo apt install ros-galactic-ackermann-msgs ros-galactic-control-toolbox
```

##### Humble

```
sudo apt install ros-humble-ackermann-msgs ros-humble-control-toolbox
```

````

File: Code/Source/VehicleDynamics/WheelControllerComponent.h (hunk 23 above)
Comment (adamdbrw) [reply Code/Source/VehicleDynamics/WheelControllerComponent.h]: For inline comments - yes

File: Code/Source/VehicleDynamics/ChassisConfiguration.cpp [hunk 27]
```diff
@@ -0,0 +1,36 @@
...
+        AxleConfiguration::Reflect(context);
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<ChassisConfiguration>()->Version(1)->Field("AxlesConfigurations", &ChassisConfiguration::m_axles);
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<ChassisConfiguration>("Chassis configuration", "Configuration of vehicle chassis")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
+                    ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC("Game"))
```
lemonade-dm: The AppearsInAddComponentMenu attribute is applicable only for Components.
For non-component types, this attribute isn't needed.

Code/Source/RobotControl/ControlConfiguration.h
```diff
@@ -19,7 +18,7 @@ namespace ROS2
     //! Configuration for handling of robot control buses.
     //! Used through ROS2RobotControlComponent.
     struct ControlConfiguration
-    {
+    { // TODO - this is two things - steering type and subscription configuration. Divide
```
spham-amzn: Would it be better to divide this struct now vs. later? in case the changes will cascade too much

Code/Source/VehicleDynamics/DriveModels/PidConfiguration.h
```diff
@@ -0,0 +1,43 @@
...
+        //! Get PID controller.
+        //! @param error Value of difference between target and state since last call
+        //! @param deltaTimeNanoseconds change in time since last call (nanoseconds)
+        //! @returns Value of computed command
+        double ComputeCommand(double error, uint64_t deltaTimeNanoseconds);
+
+    private:
+        double m_p = 1.0;
+        double m_i = 0.0;
+        double m_d = 0.0;
```
pawellech1: Consider longer variable names instead of using single letters. I think it would be more readable.
```suggestion
        double m_proportionalGain = 1.0;
        double m_integralGain = 0.0;
        double m_derivativeGain = 0.0;
```

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+    // TODO - speed and steering handling is quite similar, possible to refactor?
+    void SimplifiedDriveModel::ApplySteering(float steering, uint64_t nsDt)
+    {
+        if (m_steeringData.empty())
+        {
+            AZ_Warning("ApplySteering", false, "Cannot apply steering since no steering elements are defined in the model");
+            return;
+        }
+
+        const double nsDtSec = double(nsDt) / 1e9;
```
pawellech1: consider renaming this variable as well as the `nsDt` function argument

Code/Source/VehicleDynamics/VehicleInputsState.h
```diff
@@ -0,0 +1,66 @@
...
+        {
+            m_input = updatedInput;
+            m_lastUpdateUs = GetTimeSinceStartupUs();
+        }
+
+        T GetValue() const
+        {
+            if (auto timeSinceStartUs = GetTimeSinceStartupUs(); timeSinceStartUs - m_lastUpdateUs > m_timeoutUs)
+            {
+                return 0;
```
pawellech1, line 44: If 0 can be a valid value of an input maybe returning `std::pair<bool, T>` would be a better solution here. result.first would indicate if the value should be considered up to date.

Code/Source/VehicleDynamics/ChassisConfiguration.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include "AxleConfiguration.h"
+#include <AzCore/Serialization/SerializeContext.h>
+#include <AzCore/std/containers/vector.h>
+#include <AzFramework/Physics/Material/PhysicsMaterial.h>
+
+namespace VehicleDynamics
+{
+    //! Drive and steering configuration for a vehicle
```
spham-amzn: Is this the right description for 'Chassis' ?

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+        for (const auto& steeringElementData : m_steeringData)
+        {
+            auto steeringEntity = steeringElementData.m_steeringEntity;
+            AZ::Vector3 currentSteeringElementRotation;
+            AZ::TransformBus::EventResult(currentSteeringElementRotation, steeringEntity, &AZ::TransformBus::Events::GetLocalRotation);
+            auto currentSteeringAngle = currentSteeringElementRotation.Dot(steeringElementData.m_turnAxis);
+            double pidCommand = m_steeringPid.ComputeCommand(steering - currentSteeringAngle, nsDt);
+            if (AZ::IsClose(pidCommand, 0.0)) // TODO - use the third argument with some reasonable value which means "close enough"
```
spham-amzn, line 89: Maybe we can move ```0.0``` out to a constexpr constant double variable in the header file so it doesnt get lost here

Code/Source/VehicleDynamics/ManualControlEventHandler.h
```diff
@@ -0,0 +1,93 @@
...
+    //! A handler for a single input event.
+    class ManualControlSingleEventHandler : private StartingPointInput::InputEventNotificationBus::Handler
+    {
+    public:
+        using OnHeldHandlerFunction = std::function<void(float)>;
+
+        //! Construct the event handler.
+        //! @param eventName which event to handle (e.g. "steering", "accelerate")
+        //! @param handler a function which handles the input, typically through re-publishing it to a vehicle input bus.
+        ManualControlSingleEventHandler(AZStd::string eventName, OnHeldHandlerFunction handler)
```
spham-amzn: minor: This could be a const ref
```suggestion
        ManualControlSingleEventHandler(const AZStd::string& eventName, OnHeldHandlerFunction handler)
```
so no need to do the std::move below

Code/Source/VehicleDynamics/VehicleModelLimits.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<VehicleModelLimits>("Vehicle Model Limits", "Limitations of speed, steering angles and other values")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "")
+                    ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC("Game"))
+                    ->Attribute(AZ::Edit::Attributes::Category, "ROS2")
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &VehicleModelLimits::m_speedLimit, "Speed Limit", "Max linear speed (mps)")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default, &VehicleModelLimits::m_steeringLimit, "Speed Limit", "Max steering angle (rad)");
```
michalpelka: from @pawelbudziszewski :
```suggestion
               AZ::Edit::UIHandlers::Default, &VehicleModelLimits::m_steeringLimit, "Steering Limit", "Max steering angle (rad)");
```

Code/Source/RobotControl/Ackermann/AckermannBus.h
adamdbrw: I updated the documentation as you suggest (comment). Overall this is a typical short name in robotics community.

Code/ros2_files.cmake
adamdbrw: The old one got corrected by clang formatting (perhaps overzealous to apply to .cmake). It is ok either way

README.md
adamdbrw: I guess it is more convenient for the user to just copy and paste it, but also if we support even more distributions there is a lot of repetition here. As long as we do not need different steps for different..

I added a handy command and additional explaination

Code/Source/RobotControl/ControlConfiguration.h
adamdbrw: I added an issue for this since a good solution would also affect several other parts. This is good to do before Manipulation since that would be another subscription and beneficial for reuse.

The issue is here: https://github.com/RobotecAI/o3de-ros2-gem/issues/230.

I removed the comment.

Code/Source/VehicleDynamics/DriveModels/PidConfiguration.h
adamdbrw: These are typical variable names for PID controllers. I added documentation comments however, on top of already existing user-side descriptions in Reflections.

Code/Source/VehicleDynamics/VehicleInputsState.h
adamdbrw: This could be the case if we had a controller that behaves differently in such cases, but for us a zero is a zero at the moment.
Note that when implementing a "sleeping" or "handbrake" behavior, this part of the code might be affected.

Code/Source/VehicleDynamics/ChassisConfiguration.h
adamdbrw: I changed the class name and naming overall to vehicle configuration

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
adamdbrw: Here, pidCommand will always be compared to a zero (we are looking whether there is anything to apply).
The third argument of AZ::IsClose (which is unused, default for now) is something we might be interested in changing.

adamdbrw:
Comments have been applied. The PR is ready for another round!

michalpelka (approved):

Code/Source/VehicleDynamics/ManualControlEventHandler.h
lemonade-dm: @spham-amzn Actually the const ref is a pessimization in this case.
At worse by accepting the parameter by value and moving, supplying a lvalue AZStd::string would perform 1 copy and 1 move.

With a `const AZStd::string&`, supplying a string literal would perform 2 copies. One to construct the temporary AZStd::string that binds to a `const AZStd::string&`, the second to then copy into the `m_eventName` member.

adamdbrw: I guess since it is a constructor that gets called 2 times optimization is not that much of an issue here.
Here, we will likely run into SSO (short string optimization) since our strings are short event names: "steering", "accelerate"

piotr-zyskowski-rai (approved):

docs/guides/ros2-gem.md
```diff
@@ -22,71 +21,140 @@ You can browse Doxygen-generated documentation on [Gem's GitHub page](https://ro
   - ROS2IMUComponent
   - ROS2LidarComponent
 - __Robot control__
-  - ROS2RobotControlComponent
+    - ROS2RobotControlComponent
```
j-rivero: NitPick: reduce spacing

Code/Source/RobotControl/Ackermann/AckermannBus.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace ROS2
+{
+    //! Interface class for handling Ackermann kinematics steering commands through EBus notifications.
+    //! The interface serves to enable control through AckermannDrive (and AckermannDriveStamped) messages.
+    class AckermannNotifications : public AZ::EBusTraits
+    {
+    public:
+        static const AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
+        static const AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
```
j-rivero: [According to the documentation](https://docs.aws.amazon.com/lumberyard/latest/apireference/struct_a_z_1_1_e_bus_traits.html#ab8ff1b1078a328a4adc53e009d55e68a), this seems to be the default. Not sure if we have a good reason to make the Single explicit.

Code/Source/RobotControl/Ackermann/AckermannBus.cpp
```diff
@@ -0,0 +1,27 @@
...
+ *
+ */
+
+#include "RobotControl/Ackermann/AckermannBus.h"
+
+namespace ROS2
+{
+    void AckermannNotificationHandler::AckermannReceived(const AckermannCommandStruct& acs)
+    {
+        Call(FN_AckermannReceived, acs);
```
j-rivero: Out of curiosity, what is this `Call` function doing?

docs/guides/ros2-gem.md
```diff
@@ -22,71 +21,140 @@ You can browse Doxygen-generated documentation on [Gem's GitHub page](https://ro
...
+#### Vehicle Model
+
+`VehicleModelComponent` serves the purpose of converting inputs such as target velocity, steering or acceleration to
+physical forces on parts of a vehicle (robot).
+`VehicleModel` has a `VehicleConfiguration` which is used to define axles, parametrize and assign wheels.
+
+The model requires a `WheelControllerComponent` present in each wheel entity. It also uses an implementation
+of `DriveModel`, which converts vehicle inputs to forces acting on steering elements and wheels.
+
+**Simplified drive model**
```
j-rivero: Nitpick: using bold format in these kind of subsections ends up rendering the same than `####` in my browser so it is hard to know that they belong to "Vehicle Model".

Code/Source/RobotControl/Twist/TwistSubscriptionHandler.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "TwistBus.h"
+#include "Utilities/ROS2Conversions.h"
+
+namespace ROS2
+{
+    void TwistSubscriptionHandler::BroadcastBus(const geometry_msgs::msg::Twist& message)
+    {
+        const AZ::Vector3 linearVelocity = ROS2Conversions::FromROS2Vector3(message.linear);
+        const AZ::Vector3 angularVelocity = ROS2Conversions::FromROS2Vector3(message.angular);
+        TwistNotificationBus::Broadcast(&TwistNotifications::TwistReceived, linearVelocity, angularVelocity);
```
j-rivero, line 19: It is unlikely that the `Twist` message is going to vary its members but might be an option to use an struct to encapsulate the linear/angular, like it was done for `Ackermann`. Not sure about the implications of this in the Broadcast call though.

Code/Source/VehicleDynamics/Utilities.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include "AxleConfiguration.h"
+#include "VehicleConfiguration.h"
+#include "WheelDynamicsData.h"
+#include <AzCore/Component/EntityId.h>
+#include <AzCore/std/containers/vector.h>
+#include <AzCore/std/string/string.h>
+
+namespace VehicleDynamics::Utilities
```
j-rivero, line 17: documentation needed for functions in this file

Code/Source/VehicleDynamics/Utilities.h
```diff
@@ -0,0 +1,25 @@
...
+{
+    AxleConfiguration Create2WheelAxle(
```
j-rivero, line 27: Don't see this functions being used in the code, if I'm not wrong. I assume that they are helpers we want to provide user with.

Code/Source/VehicleDynamics/VehicleConfiguration.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "AxleConfiguration.h"
+#include <AzCore/Serialization/SerializeContext.h>
+#include <AzCore/std/containers/vector.h>
+#include <AzFramework/Physics/Material/PhysicsMaterial.h>
+
+namespace VehicleDynamics
+{
+    //! Drive and steering configuration for a vehicle
+    class VehicleConfiguration
```
j-rivero, line 18: Seems weird that we have no reference in a class called "VehicleConfiguration" to a possible chassis, a total mass, etc, just axles.

Code/Source/VehicleDynamics/VehicleInputControlBus.h
```diff
@@ -0,0 +1,56 @@
...
+    //! (Use cruise system to set cruise speed).
+    class VehicleInputControlRequests
+    {
+    public:
+        AZ_RTTI(VehicleInputControlRequests, "{AB0F1D2A-3A73-41B6-B882-62316DE32010}");
+        virtual ~VehicleInputControlRequests() = default;
+
+        //! Set target for the vehicle linear speed. It should be realized over time according to drive model.
+        //! @param speedMps is a linear speed in meters per second with the plus sign in the forward direction.
+        virtual void SetTargetLinearSpeed(float speedMps) = 0;
```
j-rivero, line 26: Nitpick: Mps can mean meters per second or miles per second :wink:

Code/Source/VehicleDynamics/VehicleInputControlBus.h
```diff
@@ -0,0 +1,56 @@
...
+
+        //! Accelerate without target speed.
+        //! @param acceleration is relative to limits of possible acceleration.
+        //! 1.0 - accelerate as much as possible, -1.0 - brake as much as possible.
+        virtual void SetTargetAcceleration(float acceleration) = 0;
```
j-rivero: Maybe use a different name for acceleration since someone would expect to send a m/s2 magnitude into a variable of this name, even if documentation says otherwise.

Code/Source/VehicleDynamics/VehicleModelComponent.h
```diff
@@ -0,0 +1,53 @@
...
+
+        void Activate() override;
+        void Deactivate() override;
+        static void GetProvidedServices(AZ::ComponentDescriptor::DependencyArrayType& provided);
+        static void Reflect(AZ::ReflectContext* context);
+
+    private:
+        void OnTick(float deltaTime, AZ::ScriptTimePoint time) override;
+
+        void SetTargetLinearSpeed(float speedMps) override;
```
j-rivero, line 41: It would be great to have documentation for these ones.

j-rivero (commented):
Did not find a good place to report this: in the class diagram we have "ChassisConfiguration" and  in the code I think it is named "VehicleConfiguration". 

Left some minor comments and questions. Overall, as far as my understanding reaches, looks great. It is pretty hard to review such a large PR correctly, sorry if I left something behind.

Code/Source/VehicleDynamics/Utilities.h
adamdbrw: This is indeed the case - I will document Utilities

Code/Source/VehicleDynamics/VehicleConfiguration.h
adamdbrw: We are only using axles at the moment. Other class members would be added as features as we extend the model.

Code/Source/RobotControl/Ackermann/AckermannBus.cpp
adamdbrw: After considering things a while I removed two classes (AckermannNotificationHandler, TwistNotificationHandler) - they were there to expose calls for LUA scripting (through Call) but it is actually better to handle this through Script Canvas.

docs/guides/ros2-gem.md
adamdbrw: Turns out I did not use all the sizes. Taking away one `#` from each header made space for `####` ones

Code/Source/RobotControl/Twist/TwistSubscriptionHandler.cpp
adamdbrw: For this PR, I am leaving it as it is. If we need something else as a part of a TwistNotification, this is going into a struct.

Code/Source/VehicleDynamics/VehicleConfiguration.h
adamdbrw: I added a comment about this

Code/Source/VehicleDynamics/VehicleInputControlBus.h
adamdbrw: These would be some serious speeds indeed! :D 
I guess if we also handled rockets..
Fortunately, there is a comment here.

Code/Source/VehicleDynamics/VehicleInputControlBus.h
adamdbrw: We have SetTarget*Fraction functions, specifically to signal arguments are -1 to 1 (relative to what model can achieve).
I renamed this function to be in line with the rest.

Code/Source/VehicleDynamics/VehicleModelComponent.h
adamdbrw: These are documented already in the interface (VehicleInputControlBus)

adamdbrw: I added a reference for this one

adamdbrw:
I addressed all the comments. I also added some Editor flair (min/max for user inputs on things like steering angle limit etc.) and removed one unnecessary field.
Diagrams have been updated.

adamdbrw:
@j-rivero everything should be addressed, you are welcome to take a look and comment / approve. We would like to merge it today.

pijaro (approved):

j-rivero:
> @j-rivero everything should be addressed, you are welcome to take a look and comment / approve.

I just read about each of the comments, thanks for taking care Adam. Fixes were just fine. Nice PR.

PR #172, opened 2022-09-13
Robot spawner

pawellech1:
ROS2SpawnerComponent created. It allows spawning robots in a specific place using services from gazebo_msgs.

User has to manually add spawnable file to the Spawner's spawnables list in the Editor before running the simulation.

### DIFF
--- a/Code/CMakeLists.txt
+++ b/Code/CMakeLists.txt
@@ -35,7 +35,7 @@ ly_add_target(
             Gem::Atom_Component_DebugCamera.Static
 )
 
-target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros)
+target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros o3de_spawning_interface_srvs )
 
 ly_add_target(
     NAME ROS2.API HEADERONLY
--- a/Code/CMakeLists.txt
+++ b/Code/CMakeLists.txt
@@ -35,7 +35,7 @@ ly_add_target(
             Gem::Atom_Component_DebugCamera.Static
 )
 
-target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros o3de_spawning_interface_srvs )
+target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros gazebo_msgs)
 
 ly_add_target(
     NAME ROS2.API HEADERONLY
--- a/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -7,10 +7,10 @@
  */
 
 #include "ROS2SpawnerComponent.h"
-#include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Math/Quaternion.h>
-#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+#include <AzCore/Serialization/EditContext.h>
 #include <AzFramework/Components/TransformComponent.h>
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
 
 namespace ROS2
 {
@@ -19,14 +19,11 @@ namespace ROS2
         auto ros2Node = ROS2Interface::Get()->GetNode();
 
         m_get_names_service = ros2Node->create_service<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames>(
-                "get_available_spawnable_names",
-                std::bind( &ROS2SpawnerComponent::GetAvailableSpawnableNames, this, AZStd::placeholders::_1, AZStd::placeholders::_2 )
-        );
+            "get_available_spawnable_names",
+            std::bind(&ROS2SpawnerComponent::GetAvailableSpawnableNames, this, AZStd::placeholders::_1, AZStd::placeholders::_2));
 
         m_spawn_service = ros2Node->create_service<o3de_spawning_interface_srvs::srv::SpawnRobot>(
-                "spawn_robot",
-                std::bind( &ROS2SpawnerComponent::SpawnRobot, this, AZStd::placeholders::_1, AZStd::placeholders::_2 )
-        );
+            "spawn_robot", std::bind(&ROS2SpawnerComponent::SpawnRobot, this, AZStd::placeholders::_1, AZStd::placeholders::_2));
     }
 
     void ROS2SpawnerComponent::Deactivate()
@@ -37,62 +34,67 @@ namespace ROS2
     {
         if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
         {
-            serialize->Class<ROS2SpawnerComponent, AZ::Component>()->Version(1)
-                ->Field("Spawner", &ROS2SpawnerComponent::m_spawnables);
+            serialize->Class<ROS2SpawnerComponent, AZ::Component>()->Version(1)->Field("Spawner", &ROS2SpawnerComponent::m_spawnables);
 
             if (AZ::EditContext* ec = serialize->GetEditContext())
             {
                 ec->Class<ROS2SpawnerComponent>("ROS2 Spawner", "Spawner component")
-                        ->ClassElement(AZ::Edit::ClassElements::EditorData, "Allows spawning")
-                        ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC("Game"))
-                        ->DataElement(
-                                AZ::Edit::UIHandlers::EntityId,
-                                &ROS2SpawnerComponent::m_spawnables,
-                                "Spawnable",
-                                "Spawnable");
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "Allows spawning")
+                    ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC("Game"))
+                    ->DataElement(AZ::Edit::UIHandlers::EntityId, &ROS2SpawnerComponent::m_spawnables, "Spawnable", "Spawnable");
             }
         }
     }
 
-    void ROS2SpawnerComponent::GetAvailableSpawnableNames(const std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Request> request, std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Response> response)
+    void ROS2SpawnerComponent::GetAvailableSpawnableNames(
+        const std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Request> request,
+        std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Response> response)
     {
-        for( const auto& spawnable : m_spawnables )
+        for (const auto& spawnable : m_spawnables)
         {
-            response->names.emplace_back( spawnable.GetHint().c_str() );
+            response->names.emplace_back(spawnable.GetHint().c_str());
         }
     }
 
-    void ROS2SpawnerComponent::SpawnRobot(const std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Request> request, std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Response> response)
+    void ROS2SpawnerComponent::SpawnRobot(
+        const std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Request> request,
+        std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Response> response)
     {
-        auto key = AZStd::string(request->robot_name.c_str(), request->robot_name.size() );
+        auto key = AZStd::string(request->robot_name.c_str(), request->robot_name.size());
 
-        auto spawnable = std::find_if( m_spawnables.begin(), m_spawnables.end(), [key](auto spwn) { return spwn.GetHint() == key; } );
-
-        if(  spawnable != m_spawnables.end() )
-        {
-            if( m_tickets.find( key ) == m_tickets.end() )
+        auto spawnable = std::find_if(
+            m_spawnables.begin(),
+            m_spawnables.end(),
+            [key](auto spwn)
             {
-                // if a ticket for this spawnable was not created but the spawnable name is correct, create the ticket and then use it to spawn an entity
-                m_tickets.insert( { spawnable->GetHint(), AzFramework::EntitySpawnTicket(*spawnable) });
+                return spwn.GetHint() == key;
+            });
+
+        if (spawnable != m_spawnables.end())
+        {
+            if (m_tickets.find(key) == m_tickets.end())
+            {
+                // if a ticket for this spawnable was not created but the spawnable name is correct, create the ticket and then use it to
+                // spawn an entity
+                m_tickets.insert({ spawnable->GetHint(), AzFramework::EntitySpawnTicket(*spawnable) });
             }
 
             auto spawner = AZ::Interface<AzFramework::SpawnableEntitiesDefinition>::Get();
 
             AzFramework::SpawnAllEntitiesOptionalArgs optionalArgs;
 
-            optionalArgs.m_preInsertionCallback = [this, position = request->position](auto id, auto view){
+            optionalArgs.m_preInsertionCallback = [this, position = request->position](auto id, auto view)
+            {
                 this->pre_spawn(
-                        id,
-                        view,
-                        AZ::Transform(
-                            AZ::Vector3(position.position.x, position.position.y, position.position.z),
-                            AZ::Quaternion(position.orientation.x, position.orientation.y, position.orientation.z, position.orientation.w),
-                            1.0f
-                        )
-                );
+                    id,
+                    view,
+                    AZ::Transform(
+                        AZ::Vector3(position.position.x, position.position.y, position.position.z),
+                        AZ::Quaternion(position.orientation.x, position.orientation.y, position.orientation.z, position.orientation.w),
+                        1.0f));
             };
 
-            spawner->SpawnAllEntities( m_tickets.at(key), optionalArgs );
+            spawner->SpawnAllEntities(m_tickets.at(key), optionalArgs);
 
             response->result = true;
             return;
@@ -101,13 +103,14 @@ namespace ROS2
         response->result = false;
     }
 
-    void ROS2SpawnerComponent::pre_spawn(AzFramework::EntitySpawnTicket::Id id, AzFramework::SpawnableEntityContainerView view, const AZ::Transform& transform)
+    void ROS2SpawnerComponent::pre_spawn(
+        AzFramework::EntitySpawnTicket::Id id, AzFramework::SpawnableEntityContainerView view, const AZ::Transform& transform)
     {
         AZ::Entity* root = *view.begin();
 
         auto* transformInterface_ = root->FindComponent<AzFramework::TransformComponent>();
 
-        transformInterface_->SetWorldTM( transform );
+        transformInterface_->SetWorldTM(transform);
     }
 
 } // namespace ROS2
--- a/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -47,28 +47,19 @@ namespace ROS2
     }
 
     void ROS2SpawnerComponent::GetAvailableSpawnableNames(
-        const std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Request> request,
-        std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Response> response)
+        const GetAvailableSpawnableNamesRequest request, GetAvailableSpawnableNamesResponse response)
     {
         for (const auto& spawnable : m_spawnables)
         {
-            response->names.emplace_back(spawnable.GetHint().c_str());
+            response->names.emplace_back(spawnable.first.c_str());
         }
     }
 
-    void ROS2SpawnerComponent::SpawnRobot(
-        const std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Request> request,
-        std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Response> response)
+    void ROS2SpawnerComponent::SpawnRobot(const SpawnRobotRequest request, SpawnRobotResponse response)
     {
-        auto key = AZStd::string(request->robot_name.c_str(), request->robot_name.size());
+        const AZStd::string key(request->robot_name.c_str(), request->robot_name.size());
 
-        auto spawnable = std::find_if(
-            m_spawnables.begin(),
-            m_spawnables.end(),
-            [key](auto spwn)
-            {
-                return spwn.GetHint() == key;
-            });
+        auto spawnable = m_spawnables.find(key);
 
         if (spawnable != m_spawnables.end())
         {
@@ -76,7 +67,7 @@ namespace ROS2
             {
                 // if a ticket for this spawnable was not created but the spawnable name is correct, create the ticket and then use it to
                 // spawn an entity
-                m_tickets.insert({ spawnable->GetHint(), AzFramework::EntitySpawnTicket(*spawnable) });
+                m_tickets.emplace(spawnable->first, AzFramework::EntitySpawnTicket(spawnable->second));
             }
 
             auto spawner = AZ::Interface<AzFramework::SpawnableEntitiesDefinition>::Get();
--- a/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "ROS2SpawnerComponent.h"
+#include <AzCore/Serialization/EditContext.h>
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+
+namespace ROS2
+{
+    void ROS2SpawnerComponent::Activate()
+    {
+        auto ros2Node = ROS2Interface::Get()->GetNode();
+
+        get_names_service = ros2Node->create_service<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames>(
+                "get_available_spawnable_names",
+                std::bind( &ROS2SpawnerComponent::GetAvailableSpawnableNames, this, AZStd::placeholders::_1, AZStd::placeholders::_2 )
+                );
+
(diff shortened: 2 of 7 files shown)


Code/Source/Spawner/ROS2SpawnerComponent.h
```diff
@@ -0,0 +1,55 @@
...
+        void Activate() override;
+
+        void Deactivate() override;
+
+        // Required Reflect function.
+        static void Reflect(AZ::ReflectContext* context);
+
+    private:
+        std::map<AZStd::string, AzFramework::EntitySpawnTicket> m_tickets = {};
+        AZStd::vector<AZ::Data::Asset<AzFramework::Spawnable>> m_spawnables = {};
```
michalpelka: no need for {}. Why not use AZStd::unordered_map here?

Code/Source/Spawner/ROS2SpawnerComponent.h
```diff
@@ -0,0 +1,55 @@
...
+
+        void Activate() override;
+
+        void Deactivate() override;
+
+        // Required Reflect function.
+        static void Reflect(AZ::ReflectContext* context);
+
+    private:
+        std::map<AZStd::string, AzFramework::EntitySpawnTicket> m_tickets = {};
```
michalpelka: no need for {}. Also switch to AZStd::map or AZStd::unodered_map.

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,116 @@
...
+                ec->Class<ROS2SpawnerComponent>("ROS2 Spawner", "Spawner component")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "Allows spawning")
+                    ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC("Game"))
+                    ->DataElement(AZ::Edit::UIHandlers::EntityId, &ROS2SpawnerComponent::m_spawnables, "Spawnable", "Spawnable");
+            }
+        }
+    }
+
+    void ROS2SpawnerComponent::GetAvailableSpawnableNames(
+        const std::shared_ptr<o3de_spawning_interface_srvs::srv::GetAvailableSpawnableNames::Request> request,
```
michalpelka: I would consider to alias this type.

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,116 @@
...
+        {
+            response->names.emplace_back(spawnable.GetHint().c_str());
+        }
+    }
+
+    void ROS2SpawnerComponent::SpawnRobot(
+        const std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Request> request,
+        std::shared_ptr<o3de_spawning_interface_srvs::srv::SpawnRobot::Response> response)
+    {
+        auto key = AZStd::string(request->robot_name.c_str(), request->robot_name.size());
```
michalpelka: const AZStd::string key (request->robot_name.c_str(), request->robot_name.size());

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,116 @@
...
+                return spwn.GetHint() == key;
+            });
+
+        if (spawnable != m_spawnables.end())
+        {
+            if (m_tickets.find(key) == m_tickets.end())
+            {
+                // if a ticket for this spawnable was not created but the spawnable name is correct, create the ticket and then use it to
+                // spawn an entity
+                m_tickets.insert({ spawnable->GetHint(), AzFramework::EntitySpawnTicket(*spawnable) });
```
michalpelka: consider using emplace

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,116 @@
...
+        AzFramework::EntitySpawnTicket::Id id, AzFramework::SpawnableEntityContainerView view, const AZ::Transform& transform)
+    {
+        AZ::Entity* root = *view.begin();
+
+        auto* transformInterface_ = root->FindComponent<AzFramework::TransformComponent>();
+
+        transformInterface_->SetWorldTM(transform);
+    }
+
+} // namespace ROS2
```
michalpelka: new line at the end

Code/CMakeLists.txt
```diff
@@ -29,12 +29,13 @@ ly_add_target(
...
             AZ::AzCore
             AZ::AzFramework
+            AZ::AzToolsFramework
             Gem::Atom_RPI.Public
             Gem::Atom_Feature_Common.Static
             Gem::Atom_Component_DebugCamera.Static
 )
 
-target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros)
+target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros o3de_spawning_interface_srvs )
```
michalpelka: Is o3de_spawning_interface_srvs some packet that you need in your ROS2 workspace?

michalpelka (commented):
Minor code issues + questions.

Code/CMakeLists.txt
adamdbrw: Is it possible to build this package within as a cmake custom command? (To invoke colcon build)

adamdbrw: Lets discuss tomorrow

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "ROS2SpawnerComponent.h"
+#include <AzCore/Math/Quaternion.h>
+#include <AzCore/Serialization/EditContext.h>
+#include <AzFramework/Components/TransformComponent.h>
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
```
adamdbrw: Is this include necessary? If not , also remove dependency from cmake

pawellech1: The Field in line `serialize->Class<ROS2SpawnerComponent, AZ::Component>()->Version(1)->Field("Spawner", &ROS2SpawnerComponent::m_spawnables);` does not work without this include.

Code/CMakeLists.txt
pawellech1: @michalpelka o3de_spawning_interface_srvs is a package in which both services used by the spawner are defined.

adamdbrw:
@pawellech1 please use "Resolve conversation" for everything you have applied, rebase on the new development and tag me for review when ready.

pawellech1:
@adamdbrw PR is ready for review

Code/Source/Spawner/ROS2SpawnerComponent.cpp
adamdbrw: It is not related - include a proper header instead (typically for classes where you Reflect):
```
#include <AzCore/Serialization/EditContext.h>
#include <AzCore/Serialization/EditContextConstants.inl>
#include <AzCore/Serialization/SerializeContext.h>
```
Probably the second one is missing

Code/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ ly_add_target(
             Gem::Atom_Component_DebugCamera.Static
 )
 
-target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros)
+target_depends_on_ros2_packages(ROS2.Static rclcpp builtin_interfaces std_msgs sensor_msgs urdfdom tf2_ros gazebo_msgs)
```
adamdbrw, line 38: We need to put this as a dependency into the README file as well

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+        m_get_names_service = ros2Node->create_service<gazebo_msgs::srv::GetWorldProperties>(
+            "get_available_spawnable_names",
+            std::bind(&ROS2SpawnerComponent::GetAvailableSpawnableNames, this, AZStd::placeholders::_1, AZStd::placeholders::_2));
+
+        m_spawn_service = ros2Node->create_service<gazebo_msgs::srv::SpawnEntity>(
+            "spawn_entity", std::bind(&ROS2SpawnerComponent::SpawnRobot, this, AZStd::placeholders::_1, AZStd::placeholders::_2));
+    }
+
+    void ROS2SpawnerComponent::Deactivate()
+    {
```
adamdbrw, line 38: It would be best if the service was not operational when the component is Deactivated - you can reset services.

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+
+namespace ROS2
+{
+    void ROS2SpawnerComponent::Activate()
+    {
+        auto ros2Node = ROS2Interface::Get()->GetNode();
+
+        m_get_names_service = ros2Node->create_service<gazebo_msgs::srv::GetWorldProperties>(
+            "get_available_spawnable_names",
```
adamdbrw, line 23: It is a good idea to include a configurable namespace here - does not cost much and adds support for multiple simulations (on network), or multiple service providers.

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    void ROS2SpawnerComponent::Deactivate()
+    {
+    }
+
+    void ROS2SpawnerComponent::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<ROS2SpawnerComponent, AZ::Component>()->Version(1)->Field("Spawner", &ROS2SpawnerComponent::m_spawnables);
```
adamdbrw: Field name should match the serialized field

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<ROS2SpawnerComponent>("ROS2 Spawner", "Spawner component")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "Allows spawning")
```
adamdbrw: extend description a bit - this should be human readable
"Facilitates spawning on robots in configurable locations" or something like this

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    void ROS2SpawnerComponent::SpawnRobot(const SpawnEntityRequest request, SpawnEntityResponse response)
+    {
+        const AZStd::string key(request->name.c_str(), request->name.size());
+
+        auto spawnable = m_spawnables.find(key);
+
+        if (spawnable != m_spawnables.end())
```
adamdbrw: Better to do an early return 
if (spawnable == m_spawnables.end())
{
  response.success = false;
  return;
}

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+                this->pre_spawn(
+                    id,
+                    view,
+                    AZ::Transform(
+                        AZ::Vector3(position.position.x, position.position.y, position.position.z),
+                        AZ::Quaternion(position.orientation.x, position.orientation.y, position.orientation.z, position.orientation.w),
+                        1.0f));
+            };
+
+            spawner->SpawnAllEntities(m_tickets.at(key), optionalArgs);
```
adamdbrw: Is this always successful? How can we learn if this succeeded? We should return success only when we are quite confident. Any other checks to determine that which could be useful here?

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+            response->success = true;
+            return;
+        }
+
+        response->success = false;
+    }
+
+    void ROS2SpawnerComponent::pre_spawn(
+        AzFramework::EntitySpawnTicket::Id id, AzFramework::SpawnableEntityContainerView view, const AZ::Transform& transform)
```
adamdbrw: Unused argument - id (use [[maybe_unused]] if it is as planned)

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        AZ::Entity* root = *view.begin();
+
+        auto* transformInterface_ = root->FindComponent<AzFramework::TransformComponent>();
+
+        transformInterface_->SetWorldTM(transform);
```
adamdbrw, line 123: Is it feasible to use TransformBus here instead? It seems the Component only works when activated, so this would be better.

Code/Source/Spawner/ROS2SpawnerComponent.h
```diff
@@ -0,0 +1,56 @@
...
+#include <gazebo_msgs/srv/spawn_entity.hpp>
+
+namespace ROS2
+{
+    using GetAvailableSpawnableNamesRequest = std::shared_ptr<gazebo_msgs::srv::GetWorldProperties::Request>;
+    using GetAvailableSpawnableNamesResponse = std::shared_ptr<gazebo_msgs::srv::GetWorldProperties::Response>;
+    using SpawnEntityRequest = std::shared_ptr<gazebo_msgs::srv::SpawnEntity::Request>;
+    using SpawnEntityResponse = std::shared_ptr<gazebo_msgs::srv::SpawnEntity::Response>;
+
+    //! TODO: doc
```
adamdbrw: Include doxygen documentation in this PR already - it is best to have it from start even with a minimal implementation

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        auto spawnable = m_spawnables.find(key);
+
+        if (spawnable != m_spawnables.end())
+        {
+            if (m_tickets.find(key) == m_tickets.end())
+            {
+                // if a ticket for this spawnable was not created but the spawnable name is correct, create the ticket and then use it to
+                // spawn an entity
+                m_tickets.emplace(spawnable->first, AzFramework::EntitySpawnTicket(spawnable->second));
```
adamdbrw: We should probably validate the name somehow - create a task for it (with lower priority)

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: The include missing from `AzToolsFramework/Entity/EditorEntityHelpers.h` is `#include <AzToolsFramework/API/EntityCompositionRequestBus.h>` but it does not remove the dependence on AzToolsFramework. I checked the includes deeper but the farthest I can get is `AzToolsFramework/UI/PropertyEditor/PropertyEditorAPI.h` which still does not solve the problem.

The dependency in the CMake is already added by another PR.

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: Should it be configurable in the editor as a component field or in some other way?

adamdbrw: For now, just add an issue for this, I guess we can start with no namespace

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: https://github.com/RobotecAI/o3de-ros2-gem/issues/205

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: https://github.com/RobotecAI/o3de-ros2-gem/issues/204

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: https://github.com/RobotecAI/o3de-ros2-gem/issues/207

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+                ec->Class<ROS2SpawnerComponent>("ROS2 Spawner", "Spawner component")
+                    ->ClassElement(AZ::Edit::ClassElements::EditorData, "Manages spawning of robots in configurable locations")
+                    ->Attribute(AZ::Edit::Attributes::AppearsInAddComponentMenu, AZ_CRC("Game"))
+                    ->DataElement(AZ::Edit::UIHandlers::EntityId, &ROS2SpawnerComponent::m_spawnables, "Spawnables", "Spawnables");
+            }
+        }
+    }
+
+    void ROS2SpawnerComponent::GetAvailableSpawnableNames(
+        const GetAvailableSpawnableNamesRequest request, GetAvailableSpawnableNamesResponse response)
```
lemonade-dm, line 60: Both of these parameters should be by reference.
The first parameter is const reference, the second with non-const reference.
```suggestion
    void ROS2SpawnerComponent::GetAvailableSpawnableNames(
        const GetAvailableSpawnableNamesRequest& request, GetAvailableSpawnableNamesResponse& response)
```

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        for (const auto& spawnable : m_spawnables)
+        {
+            response->model_names.emplace_back(spawnable.first.c_str());
+        }
+    }
+
+    void ROS2SpawnerComponent::SpawnRobot(const SpawnEntityRequest request, SpawnEntityResponse response)
```
lemonade-dm: repeat: Both parameters should be accepted by reference(const and non-const).
```suggestion
    void ROS2SpawnerComponent::SpawnRobot(const SpawnEntityRequest& request, SpawnEntityResponse& response)
```

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: ROS2 Node::create_service method does not accept the callback with reference arguments. It could work if I switched from bind:
`ros2Node->create_service<gazebo_msgs::srv::GetWorldProperties>("get_available_spawnable_names", std::bind(&ROS2SpawnerComponent::GetAvailableSpawnableNames, this, AZStd::placeholders::_1, AZStd::placeholders::_2));` 
to lamda:
`ros2Node->create_service<gazebo_msgs::srv::GetWorldProperties>("get_available_spawnable_names", [this](const GetAvailableSpawnableNamesRequest req, GetAvailableSpawnableNamesResponse res){ this->GetAvailableSpawnableNames(req, res); })`

In the second case the GetAvailableSpawnableNames signature can look the way you suggested. But I think we would not gain anything from it since the copy is still needed to call the lambda callback? Correct me please if I am wrong.

adamdbrw (approved):

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: It is also done without the references in [ROS2 service examples](https://docs.ros.org/en/galactic/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Cpp-Service-And-Client.html#write-the-service-node), so I assumed it is the correct way.

lemonade-dm: It looks like [GetAvailableSpawnableNamesResponse](https://github.com/RobotecAI/o3de-ros2-gem/pull/172/files#diff-2f072dcae025d6df3fbcdf9ab9981856ef73e0c68f4b84e770eaf4098d8ec7e7R21) is type alias to a `shared_ptr`.

I thought this was an raw value object, so it wasn't being updated.
It makes sense now that the parameter is passed by value.

Code/Source/Spawner/ROS2SpawnerComponent.h
```diff
@@ -0,0 +1,57 @@
...
+        AZStd::unordered_map<AZStd::string, AzFramework::EntitySpawnTicket> m_tickets;
+
+        rclcpp::Service<gazebo_msgs::srv::GetWorldProperties>::SharedPtr m_get_names_service;
+        rclcpp::Service<gazebo_msgs::srv::SpawnEntity>::SharedPtr m_spawn_service;
+
+        void GetAvailableSpawnableNames(const GetAvailableSpawnableNamesRequest request, GetAvailableSpawnableNamesResponse response);
+
+        void SpawnRobot(const SpawnEntityRequest request, SpawnEntityResponse response);
+
+        void pre_spawn(AzFramework::EntitySpawnTicket::Id, AzFramework::SpawnableEntityContainerView, const AZ::Transform&);
```
lemonade-dm: The naming scheme is inconsistent here. 
For changes going into the O3DE repo or O3DE-Extra repo, function names should be [UpperCamelCase](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#function-names)

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        response->success = true;
+    }
+
+    void ROS2SpawnerComponent::pre_spawn(
+        AzFramework::EntitySpawnTicket::Id id [[maybe_unused]],
+        AzFramework::SpawnableEntityContainerView view,
+        const AZ::Transform& transform)
+    {
+        AZ::Entity* root = *view.begin();
```
lemonade-dm, line 118: This should check if the view is not empty before grabbing the front instance.
Also `.front()` can be used as a better way to clarify the intent that the first element is being access as opposed to `*begin()`.

```suggestion
        if (view.empty())
        {
            return;
        }
        
        AZ::Entity* root = view.front();
```

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        for (const auto& spawnable : m_spawnables)
+        {
+            response->model_names.emplace_back(spawnable.first.c_str());
+        }
+    }
+
+    void ROS2SpawnerComponent::SpawnRobot(const SpawnEntityRequest request, SpawnEntityResponse response)
+    {
+        const AZStd::string key(request->name.c_str(), request->name.size());
```
lemonade-dm: A copy into a string isn't actually needed here.
This can be stored in a AZStd::string_view.
```suggestion
        AZStd::string_view key(request->name.c_str(), request->name.size());
```

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+        auto spawnable = m_spawnables.find(key);
+
+        if (spawnable == m_spawnables.end())
```
lumberyard-employee-dm, line 66: FYI we support the C++20 associative container [contains](https://en.cppreference.com/w/cpp/container/unordered_map/contains) method.

So this can be simplified to

```suggestion
        if (!m_spawnables.contains(key))
```

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+        {
+            response->success = false;
+            response->status_message = "Requested spawnable name not found";
+            return;
+        }
+
+        if (m_tickets.find(key) == m_tickets.end())
```
lemonade-dm: FYI we support the C++20 associative container [contains](https://en.cppreference.com/w/cpp/container/unordered_map/contains) method.

So this can be simplified to

```suggestion
        if (!m_tickets.contains(key))
```

Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -0,0 +1,113 @@
...
+    void ROS2SpawnerComponent::Activate()
+    {
+        auto ros2Node = ROS2Interface::Get()->GetNode();
+
+        m_get_names_service = ros2Node->create_service<gazebo_msgs::srv::GetWorldProperties>(
+            "get_available_spawnable_names",
+            std::bind(&ROS2SpawnerComponent::GetAvailableSpawnableNames, this, AZStd::placeholders::_1, AZStd::placeholders::_2));
+
+        m_spawn_service = ros2Node->create_service<gazebo_msgs::srv::SpawnEntity>(
+            "spawn_entity", std::bind(&ROS2SpawnerComponent::SpawnRobot, this, AZStd::placeholders::_1, AZStd::placeholders::_2));
```
lemonade-dm: It is safer to use a [lambda than std::bind](https://abseil.io/tips/108) and slightly more performant since it doesn't have to marshal the parameters to a tuple as part of binding them to unspecified std::bind result type.

```suggestion
            [this](const GetAvailableSpawnableNamesRequest request, GetAvailableSpawnableNamesResponse response) { GetAvailableSpawnableNames(request, response); });

        m_spawn_service = ros2Node->create_service<gazebo_msgs::srv::SpawnEntity>(
            "spawn_entity", [this](const SpawnEntityRequest request, SpawnEntityResponse response) { SpawnRobot(request, response); });
```

Code/Source/Spawner/ROS2SpawnerComponent.h
```diff
@@ -0,0 +1,57 @@
...
+
+        // Required Reflect function.
+        static void Reflect(AZ::ReflectContext* context);
+
+    private:
+        AZStd::unordered_map<AZStd::string, AZ::Data::Asset<AzFramework::Spawnable>> m_spawnables;
+        AZStd::unordered_map<AZStd::string, AzFramework::EntitySpawnTicket> m_tickets;
+
+        rclcpp::Service<gazebo_msgs::srv::GetWorldProperties>::SharedPtr m_get_names_service;
+        rclcpp::Service<gazebo_msgs::srv::SpawnEntity>::SharedPtr m_spawn_service;
```
lemonade-dm: member variables names should be lowerCamel case for code going into O3DE per the [Coding Standard](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#variable-names)

Code/Source/Spawner/ROS2SpawnerComponent.h
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/Component/Component.h>
+#include <AzFramework/Spawnable/Spawnable.h>
+#include <AzFramework/Spawnable/SpawnableEntitiesInterface.h>
+#include <ROS2SystemComponent.h>
```
lemonade-dm: Is the ROSSystemComponent.h include actually needed in this file?

Code/Source/Spawner/ROS2SpawnerComponent.cpp
pawellech1: I don't see a `front` method in `SpawnableEntityContainerView` and the code did not compile. I just added the `.empty()` check.

Code/Source/Spawner/ROS2SpawnerComponent.h
pawellech1: I changed it to including `<rclcpp/rclcpp.hpp>` and `<ROS2/ROS2Bus.h>` in `ROS2SpawnerComponent.h` and `ROS2SpawnerComponent.cpp` respectivly.

PR #176, opened 2022-09-13
Spawn points provider

pawellech1:
ROS2SpawnPointsProviderComponent created. It allows user to specify spawn points from Editor before the simulation. It also allows the user to get the list of specified points at runtime by using ros service.

To specify a spawn point user should create an entity and add ROS2SpawnPointsProviderComponent to it. Spawn points should be add as a child entities of this entity. The SpawnPointComponent should be add to all child entities.


adamdbrw:
@pawellech1 what is the status here? Should it be closed in favor of the newer PR?

Issue #169, opened 2022-09-09
Editor crashes after import URDF after change of level


michalpelka:
**Bug description**
The problem show itself when the user load URDF (minimal example) and tries to change level in o3de Editor.

**To Reproduce**
1. Import minimal URDF
2. Click `File` >> `Open Level` >> choose different level. In dialog click `Close without saving`
3. Observe o3de Editor crashes.

**Expected behavior**
Swift change to other level

**Test environment**
 - OS: Ubuntu 22.04
 - ROS: Galactic

**Additional context**
The problem is persistent in its nature. The Editor crashes with such scenario:
1. Import minimal URDF
2. Click `File` >> `Open Level` >> choose different level. In dialog click `Close without saving`
3. Save scene and close
4. Open editor, choose level with imported robot.
5. Click `File` >> `Open Level` >> choose different level. In dialog click `Close without saving`
4. Observe o3de Editor crashes

This issue can be linked with #164 .
**Minimal URDF:**
```
<?xml version="1.0"?>
<robot name="myfirst">
  <link name="base_link">
    <visual>
      <geometry>
        <cylinder length="0.6" radius="0.2"/>
      </geometry>
    </visual>
  </link>
</robot>
````

michalpelka:
It was fixed with  #228

Issue #164, opened 2022-09-08
URDF importer crashes on Overwrite


michalpelka:
**Bug description**
The problem manifest itself when the user tries to reload URDF and select overwrite.
**To Reproduce**
1. Go to 'Robot importer'
2. Click on 'Load'
3. Select URDF (even on minimal urdf)
4. Optionally remove model from scene
5. Go to 'Robot importer'
6. Click on 'Load'
7. Select the same URDF once again. 
8. Editor crashes

**Expected behavior**
The prefab should be rebuild and model should be changed in the scene.

**Test environment**
 - OS: Ubuntu 22.04
 - ROS: Humble
 - ROS2 Gem [state](https://github.com/RobotecAI/o3de-ros2-gem/commit/4d7d7baec5e9824e0d891a6eff7e9a526c6cde26)
 
 **Minimal URDF**:
 ```
 <robot name="test">

    <link name="base">
        <inertial>
        <origin xyz="0. 0. 0."/>
        <mass value="1."/>
        <inertia ixx="1." ixy="0." ixz="0." iyy="1." iyz="0." izz="1."/>
        </inertial>
        <visual>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <box size="1 1 1"/>
            </geometry>
        </visual>
        <collision>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <box size="1 1 1"/>
            </geometry>
        </collision>
    </link>
</robot>
```

Issue #171, opened 2022-09-13
Editor crashes on attempt to import URDF with error


michalpelka:
**Bug description**
We try to import URDF with some error (e.g. that attached).
The editor crahes

**To Reproduce**
1. Go to 'Robot Importer'
2. Click on 'Load'
3. Choose the URDF with error (e.g. joint with the same name)'
4. Message box appears saying 'Failed to parse ....'
5. After clicking OK the Editor shuts down

**Expected behavior**
It would be nice to inform user about problem with some GUI message, this crash is not beneficial.

**Screenshots / videos**
```
Module: Success!

Error:   joint 'ups_again' is not unique.
         at line 194 in /home/michal/ros2_ws/src/urdfdom/urdf_parser/src/model.cpp

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

**Test environment**
 - OS: Ubuntu 22.04
 - ROS: Humble

**Additional context**
The URDF used to trigger this behavior
https://gist.github.com/michalpelka/53daa7300c9fe5efaffca96b61590aae

michalpelka:
issue is being fixed in #188

PR #188, opened 2022-09-14
Fix Editor crashes on URDF prefab overwite and import non-valid URDF

michalpelka:


michalpelka:
That PR fixes: #164 #171, but sadly do not fix #169.

Code/Source/RobotImporter/URDF/RobotImporter.cpp
```diff
@@ -14,16 +14,19 @@ namespace ROS2
     RobotImporter::RobotImporter(LoggerFunction logger)
         : m_logger(std::move(logger))
         , m_isProcessingAssets(false)
+        , m_LoadingURDFFailed(false)
```
pijaro: nitpick
```suggestion
        , m_loadingURDFFailed(false)
```

Code/Source/RobotImporter/URDF/RobotImporter.cpp
```diff
@@ -42,21 +45,28 @@ namespace ROS2
...
+        else if (m_LoadingURDFFailed)
         {
-            auto errorMessage = AZStd::string::format("Importing robot definition failed with error: %s", outcome.GetError().c_str());
-            m_logger(LogLevel::Error, errorMessage);
             importFinishedCb();
             return;
         }
-        m_logger(LogLevel::Info, AZStd::string::format("Imported %s", m_prefabMaker->GetPrefabPath().c_str()));
-        importFinishedCb();
+        AZ_Assert(m_prefabMaker, "Prefab maker is not initialized");
```
pijaro: I believe this is an unnecessary assert. Move it to the top or remove it.

michalpelka: I do not agree - we need to assert when URDF is not failed, but m_prefab optional is empty. I changed simplified logic in this function to make this assert more relevant. Hope you like it more now.

pijaro: I was thinking about the rule in general, that we should assert that "something" is "correct" before we start using it in any way. 

But yes, I see the reason behind that implementation. In this case, I'm ok with it.

adamdbrw (approved):

PR #189, opened 2022-09-15
pid configuration and reflection

adamdbrw:

Note - not a final solution, but an enabler

Final solution - use abstraction of Drive Model

### DIFF
--- a/Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
+++ b/Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
@@ -9,15 +9,45 @@
 #include "SimplifiedDriveModel.h"
 #include "VehicleDynamics/Utilities.h"
 #include <AzCore/Component/TransformBus.h>
+#include <AzCore/Serialization/EditContext.h>
+#include <AzCore/Serialization/EditContextConstants.inl>
+#include <AzCore/Serialization/SerializeContext.h>
 #include <AzFramework/Physics/RigidBodyBus.h>
 
 namespace VehicleDynamics
 {
-    SimplifiedDriveModel::SimplifiedDriveModel()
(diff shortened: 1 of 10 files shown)


Code/Source/VehicleDynamics/VehicleModelComponent.cpp
```diff
@@ -105,7 +112,7 @@ namespace VehicleDynamics
         }
 
         uint64_t deltaTimeNs = deltaTime * 1000000000;
```
michalpelka, line 114:
```suggestion
        const uint64_t deltaTimeNs = deltaTime * 1e9;
```

michalpelka (approved):

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -29,14 +59,14 @@ namespace VehicleDynamics
     // TODO - speed and steering handling is quite similar, possible to refactor?
     void SimplifiedDriveModel::ApplySteering(float steering, const ChassisConfiguration& vehicleChassis, uint64_t nsDt)
     {
-        const double nsDtSec = double(nsDt)/1e9;
```
pawelbudziszewski, line 32: We should leave this.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -50,7 +80,7 @@ namespace VehicleDynamics
                 continue;
             }
 
-            auto torque = pidCommand * nsDtSec;
```
pawelbudziszewski, line 53: We should leave this.

pawelbudziszewski (commented):
Feature requests:

- Easy way to define max speed (the target speed when pressing up arrow) - eg. expose it to the UI
- Easy way to define wheel rotation axis

PR #197, opened 2022-09-16
URDF importer locale warning

piotr-zyskowski-rai:
Since URDF parser uses locale information, it might incorrectly read files if the decimal delimiter is not compatible. This PR fixes this by forcing numeric locale to be `en_US.UTF-8` which is the most commonly used in the URDF files.

### DIFF
--- a/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
+++ b/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
@@ -29,26 +29,6 @@ namespace ROS2
     {
         static const char* collidersMakerLoggingTag = "CollidersMaker";
 
-        void verifyTheGeometry(const urdf::Cylinder& cylinderGeometry)
-        {
-            // Due to the fact that URDF parser takes into account the locale information, incompatibility between URDF file locale and
-            // system locale might lead to incorrect URDF parsing. Mainly it affects floating point numbers, and its decimal separator. When
-            // locales are set to system with comma as decimal separator and URDF file is created with dot as decimal separator, URDF parser
-            // will trim the floating point number after comma. For example, if cylinder radius is 0.1, URDF parser will parse it as 0. At
-            // the moment of writing this code, O3DE colliders components does not deal correctly with zero radius/length, which leads to
-            // crashes. To at least point user to where the issue might be coming from, we check if the radius/length is zero, and if
-            // it is, the warning is printed
-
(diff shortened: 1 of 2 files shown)


michalpelka (approved):

adamdbrw:
This is not acceptable - this function affects the application:

"The modifications remain in effect and influences the execution of all locale-sensitive C library functions until the next call to setlocale"

If workaround is necessary (we could also use warnings / asserts for zero values with hints of the possible reason), set and restore just after parsing: you can query with https://en.cppreference.com/w/cpp/locale/locale.

adamdbrw (changes requested):
Please add a warning instead on zero value

adamdbrw:
@piotr-zyskowski-rai is it only applicable to cylinders? Or rather all primitive shapes (and both visual and colliders)

piotr-zyskowski-rai:
> @piotr-zyskowski-rai is it only applicable to cylinders? Or rather all primitive shapes (and both visual and colliders)

@adamdbrw It most likely is. I tested it and it seems, however, that a zero-sized box or sphere of radius 0.0  does not cause Editor termination.

Locale incompatibility should actually cause way more problems, for example with the positions and rotations but there is no easy way to check that the value does not make sense for each case.

Locale incompatibility should cause problems with
- position, rotation - not much we can do here since all zeros rotation or translation are perfectly valid 
- scale - zero scale does not make any sense, might be worth notifying
- primitive sizes - the same
- inertial origin - perfectly fine with zeros
- inertial mass - zeros do not make any sense
- inertial inertia - all zeros do not make  sense, individual zeros do
- joint axis - all zeros do not make sense, individual zeros do
- color - all zeros make sense

And the question is whether we want to check guard all of the ones that we can, but still, have the URDF import fail. The whole robot will probably be messed up anyway. The cylinder is different because at the moment it causes violent Editor termination which is not nice and it would be nice to have a warning that is near the end of the log file with a hint of what might be the cause.

What we might do to tackle this issue overall, is to check the current locale settings before importing the URDF and print the warning if it is not `en_US` and only additionally guard the cylinder case, since it is "destructive".

What do you think?

adamdbrw:
I agree - lets also submit an issue for cylinder to O3DE, it should not be behaving like this

adamdbrw (approved):

adamdbrw (approved):

PR #199, opened 2022-09-16
Set larger number of iterations with imported URDFs

michalpelka:
The number of iterations vof solvers are set to larger number 40 for position and 10 for velocity

### DIFF
--- a/Code/Source/RobotImporter/URDF/InertialsMaker.cpp
+++ b/Code/Source/RobotImporter/URDF/InertialsMaker.cpp
@@ -14,6 +14,9 @@
 
 namespace ROS2
 {
+    constexpr AZ::u8 kMinimalNumPosSolv = 40;
+    constexpr AZ::u8 kMinimalNumVelSolv = 10;
+
     void InertialsMaker::AddInertial(urdf::InertialSharedPtr inertial, AZ::EntityId entityId)
     {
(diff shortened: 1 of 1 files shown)


Code/Source/RobotImporter/URDF/InertialsMaker.cpp
```diff
@@ -14,6 +14,9 @@
 
 namespace ROS2
 {
+    constexpr AZ::u8 kMinimalNumPosSolv = 40;
```
adamdbrw, line 19: Comment why

Code/Source/RobotImporter/URDF/InertialsMaker.cpp
```diff
@@ -24,6 +27,10 @@ namespace ROS2
 
         AZ::Entity* entity = AzToolsFramework::GetEntityById(entityId);
         PhysX::EditorRigidBodyConfiguration rigidBodyConfiguration;
+        PhysX::RigidBodyConfiguration physxSpecificConfiguration;
```
adamdbrw, line 32: Check and only replace if defaults are lower

michalpelka:
Thanks for the last review @adamdbrw, applied all your comments.

adamdbrw (approved):

PR #203, opened 2022-09-20
Add missing include that was being added in unity builds

spham-amzn:
Fixes the following error in non-unity builds:
```
/home/github/o3de-ros2-gem/Code/Source/RobotImporter/URDF/JointsMaker.cpp:84:40: error: use of undeclared identifier 'URDF'
        const AZ::Vector3 joint_axis = URDF::TypeConversions::ConvertVector3(joint->axis);
                                       ^
1 error generated.

```


adamdbrw (approved):
Thank you Steve! These are different builds which we do not run, it is great that you are able to catch these issues.

PR #206, opened 2022-09-20
Better handling of vehicle input limits and zeroing

adamdbrw:
Please also test whether the component parameters work as expected.

### DIFF
--- a/Code/Source/VehicleDynamics/VehicleInputsState.h
+++ b/Code/Source/VehicleDynamics/VehicleInputsState.h
@@ -7,12 +7,60 @@
  */
 #pragma once
 
+#include <AzCore/Time/ITime.h>
+#include <type_traits>
+//#include <concepts> //C++20
+
 namespace VehicleDynamics
 {
+    /* C++ 20
+    template<typename T>
+    concept Zeroable = std::integral<T> or std::floating_point<T>; // arithmetic types
+
+    template<typename T>
+    requires Zeroable<T> class InputZeroedOnTimeout
+    */
+
+    //! Inputs with an expiration date - effectively is zero after a certain time since update
+    template<typename T, typename = typename std::enable_if<std::is_arithmetic<T>::value, T>::type>
+    class InputZeroedOnTimeout
+    {
+    public:
+        InputZeroedOnTimeout(int64_t timeoutUs = 200000)
+            : m_timeoutUs(timeoutUs)
+        {
+        }
+
+        void UpdateValue(T updatedInput)
+        {
+            m_input = updatedInput;
+            m_lastUpdateUs = GetTimeSinceStartupUs();
+        }
+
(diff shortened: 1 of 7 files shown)


Code/Source/VehicleDynamics/VehicleInputsState.h
```diff
@@ -7,12 +7,60 @@
...
+            return m_input;
+        }
+
+    private:
+        int64_t GetTimeSinceStartupUs() const
+        {
+            return static_cast<int64_t>(AZ::Interface<AZ::ITime>::Get()->GetElapsedTimeUs());
+        }
+
+        int64_t m_timeoutUs;
```
michalpelka, line 55:
```suggestion
     const int64_t m_timeoutUs;
```
I also do not understand the need for SFINAE in this class.

Code/Source/VehicleDynamics/VehicleModelLimits.cpp
```diff
@@ -0,0 +1,45 @@
...
+                        &VehicleModelLimits::m_steeringLimitRads,
+                        "Speed Limit",
+                        "Max steering angle (rads)");
+            }
+        }
+    }
+
+    float VehicleModelLimits::LimitValue(float value, float absoluteLimit)
+    {
+        return value >= 0 ? AZStd::min(value, absoluteLimit) : AZStd::max(value, -absoluteLimit);
```
michalpelka, line 43: Consider omitting branching:
```suggestion
return AZStd::min(AZStd::max(value, -absoluteLimit), absoluteLimit) ;
```

michalpelka (commented):
Minor things from my side. 
I did not run it yet.

Code/Source/VehicleDynamics/VehicleModelLimits.cpp
adamdbrw: While the second option works and is shorter, I believe it is a bit less intuitive to understand.

Code/Source/VehicleDynamics/VehicleInputsState.h
adamdbrw: Primary reason for including it is that it informs the developer of the intent for the templated type without having him look into implementation. 
Also, compilation errors for invalid type are different, but I guess it is up to personal preference which is easier to understand.

adamdbrw: Another issue is that some non-arithmetic types can have zero assigned, which is not something we would like (e.g. std::string).

Code/Source/VehicleDynamics/VehicleModelLimits.cpp
adamdbrw: Replaced with AZStd::clamp, which is there :)

PR #213, opened 2022-09-21
Wheel dynamics improvements:

adamdbrw:
- removed turning input negative (instead, set axis and input properly)
- wheel radius is now configurable (per axle), defaults to 0.35 meters
- drive wheels and steering elements are now computed on activation, instead of each frame
- some names have been changed to match the naming convention

### DIFF
--- a/Code/Source/VehicleDynamics/WheelDynamicsData.h
+++ b/Code/Source/VehicleDynamics/WheelDynamicsData.h
@@ -12,7 +12,6 @@ namespace VehicleDynamics
     //! Data structure to pass wheel dynamics data for a single wheel
     struct WheelDynamicsData
     {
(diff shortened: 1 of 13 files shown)


Code/Source/VehicleDynamics/WheelDynamicsData.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ */
+#pragma once
+
+namespace VehicleDynamics
+{
+    //! Data structure to pass wheel dynamics data for a single wheel
+    struct WheelDynamicsData
+    {
+    public:
```
michalpelka: superfluous public

michalpelka (approved):

PR #219, opened 2022-09-21
Removed auto-created colliders in URDF importer

michalpelka:
Since o3de lifted the requirement that an entity needs to have a collider in order to have a joint, there is no need to create dummy colliders.


adamdbrw:
Neat, less code is better!

adamdbrw (approved):

PR #239, opened 2022-09-28
Vehicle controller physics tuning

pawelbudziszewski:
Created a more elegant and universal solution: #242. 

Previous description:
```
Added:
- Possibility to limit torque impulse for speed control. Can be controlled via `Max Speed Impulse` field in the GUI. The reasonable starting point for tuning: `500.0`. Set to `0.0` to disable.
~~- Printing debug information for speed and steering (useful for PID tuning). Can be turned on/off using switches in the GUI~~
[Moved to #240 ] 

Setting torque limit allows using more aggressive PID settings (mainly P) minimising problems with stability.
```

### DIFF
--- a/Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
+++ b/Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
@@ -46,7 +46,9 @@ namespace VehicleDynamics
                         "Configuration of speed PID controller")
                     ->DataElement(AZ::Edit::UIHandlers::Default, 
                         &SimplifiedDriveModel::maxSpeedImpulse, 
-                        "Max speed impulse", "Speed torque impulse limit [0, INF]. Set to 0.0 to disable.")
+                        "Maximum wheel torque", "Maximum torque force that can be applied to wheels regardless of controller output [0, INF]."
+                        " Set to 0.0 to disable this limit.")
+                    ->Attribute(AZ::Edit::Attributes::Min, 0.0f)
                     ->DataElement(AZ::Edit::UIHandlers::Default, 
                         &SimplifiedDriveModel::showSteeringDebugInfo, "Show steering debug messages", "")
                     ->DataElement(AZ::Edit::UIHandlers::Default, 
@@ -110,7 +112,7 @@ namespace VehicleDynamics
 
             if (showSteeringDebugInfo)
             {
-                AZ_Warning("ApplySteering", false, "Steering target: %f current: %f impulse: %f", steering, currentSteeringAngle, pidCommand);
+                AZ_TracePrintf("ApplySteering", false, "Steering target: %f current: %f impulse: %f", steering, currentSteeringAngle, pidCommand);
             }            
         }
     }
@@ -150,13 +152,13 @@ namespace VehicleDynamics
                 continue;
             }
 
-            if (maxSpeedImpulse>0.0)
+            if (maxSpeedImpulse > 0.0)
             {
-                if (pidCommand>maxSpeedImpulse)
+                if (pidCommand > maxSpeedImpulse)
                 {
                     pidCommand = maxSpeedImpulse;
                 }
-                if (pidCommand<-maxSpeedImpulse)
+                if (pidCommand < -maxSpeedImpulse)
                 {
                     pidCommand = -maxSpeedImpulse;
                 }
@@ -169,7 +171,7 @@ namespace VehicleDynamics
 
             if (showSpeedDebugInfo)
             {
-                AZ_Warning("ApplySpeed", false, "Speed target: %f current: %f impulse: %f", desiredAngularSpeedX, currentAngularSpeedX, pidCommand);
+                AZ_TracePrintf("ApplySpeed", false, "Speed target: %f current: %f impulse: %f", desiredAngularSpeedX, currentAngularSpeedX, pidCommand);
             }            
         }
     }
--- a/Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
+++ b/Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
@@ -25,9 +25,7 @@ namespace VehicleDynamics
                 ->Version(1)
                 ->Field("SteeringPID", &SimplifiedDriveModel::m_steeringPid)
                 ->Field("SpeedPID", &SimplifiedDriveModel::m_speedPid)
-                ->Field("MaxSpeedImpulse", &SimplifiedDriveModel::maxSpeedImpulse)
-                ->Field("ShowSteeringDebugInfo", &SimplifiedDriveModel::showSteeringDebugInfo)
-                ->Field("ShowSpeedDebugInfo", &SimplifiedDriveModel::showSpeedDebugInfo);
+                ->Field("MaxSpeedImpulse", &SimplifiedDriveModel::maxSpeedImpulse);
 
 
             if (AZ::EditContext* ec = serialize->GetEditContext())
@@ -46,13 +44,10 @@ namespace VehicleDynamics
                         "Configuration of speed PID controller")
                     ->DataElement(AZ::Edit::UIHandlers::Default, 
                         &SimplifiedDriveModel::maxSpeedImpulse, 
-                        "Maximum wheel torque", "Maximum torque force that can be applied to wheels regardless of controller output [0, INF]."
-                        " Set to 0.0 to disable this limit.")
-                    ->Attribute(AZ::Edit::Attributes::Min, 0.0f)
-                    ->DataElement(AZ::Edit::UIHandlers::Default, 
-                        &SimplifiedDriveModel::showSteeringDebugInfo, "Show steering debug messages", "")
-                    ->DataElement(AZ::Edit::UIHandlers::Default, 
-                        &SimplifiedDriveModel::showSpeedDebugInfo, "Show speed debug messages", "");
+                        "Maximum wheel torque", 
+                        "Maximum torque force (per 1 second) that can be applied to wheels regardless of controller "
+                        "output [0, INF]. Set to 0.0 to disable this limit.")
+                    ->Attribute(AZ::Edit::Attributes::Min, 0.0f);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -39,7 +43,14 @@ namespace VehicleDynamics
                         AZ::Edit::UIHandlers::Default,
                         &SimplifiedDriveModel::m_speedPid,
                         "Speed PID",
-                        "Configuration of speed PID controller");
+                        "Configuration of speed PID controller")
+                    ->DataElement(AZ::Edit::UIHandlers::Default, 
+                        &SimplifiedDriveModel::maxSpeedImpulse, 
+                        "Max speed impulse", "Speed torque impulse limit [0, INF]. Set to 0.0 to disable.")
```
adamdbrw: Use attribute Min here (input will be limited). See VehicleLimits.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -96,6 +107,11 @@ namespace VehicleDynamics
             AZ::TransformBus::EventResult(steeringElementTransform, steeringEntity, &AZ::TransformBus::Events::GetWorldTM);
             auto transformedTorqueVector = steeringElementTransform.TransformVector(steeringElementData.m_turnAxis * torque);
             Physics::RigidBodyRequestBus::Event(steeringEntity, &Physics::RigidBodyRequests::ApplyAngularImpulse, transformedTorqueVector);
+
+            if (showSteeringDebugInfo)
+            {
+                AZ_Warning("ApplySteering", false, "Steering target: %f current: %f impulse: %f", steering, currentSteeringAngle, pidCommand);
```
adamdbrw: Use Trace Print, this is not a warning.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -134,10 +150,27 @@ namespace VehicleDynamics
                 continue;
             }
 
+            if (maxSpeedImpulse>0.0)
```
adamdbrw: reformat with clang-13 (I see no spaces between sign etc.)

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -134,10 +150,27 @@ namespace VehicleDynamics
...
+            }
+
             auto impulse = pidCommand * deltaTimeSec;
 
             auto transformedTorqueVector = wheelTransform.TransformVector(wheelData.m_driveAxis * impulse);
             Physics::RigidBodyRequestBus::Event(wheelEntity, &Physics::RigidBodyRequests::ApplyAngularImpulse, transformedTorqueVector);
+
+            if (showSpeedDebugInfo)
+            {
+                AZ_Warning("ApplySpeed", false, "Speed target: %f current: %f impulse: %f", desiredAngularSpeedX, currentAngularSpeedX, pidCommand);
```
adamdbrw: Trace print

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -24,7 +24,11 @@ namespace VehicleDynamics
             serialize->Class<SimplifiedDriveModel, DriveModel>()
                 ->Version(1)
                 ->Field("SteeringPID", &SimplifiedDriveModel::m_steeringPid)
-                ->Field("SpeedPID", &SimplifiedDriveModel::m_speedPid);
+                ->Field("SpeedPID", &SimplifiedDriveModel::m_speedPid)
+                ->Field("MaxSpeedImpulse", &SimplifiedDriveModel::maxSpeedImpulse)
```
adamdbrw: This is something that always needs tuning? There is no good default or formula based on radius, mass?

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -24,7 +24,11 @@ namespace VehicleDynamics
...
+                ->Field("ShowSteeringDebugInfo", &SimplifiedDriveModel::showSteeringDebugInfo)
```
adamdbrw: Could you please move these debug fields (and debugs themselves) to a separate PR? This is different than limiting impulse.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
adamdbrw: Maybe "Maximum torque force that can be applied to wheels regardless of controller output"

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
adamdbrw: Perhaps one is enough? Just one variable "show debug values". Perhaps actually showing them in a component fields (e.g. so that the user does not need to dig in the logs) would be good? 

We need to have 3 fields for speed and 3 for steering, a nice layout of 3x2 would be good to present values.

That is, if you believe it would be useful for users in general. If it is only for us, then move it out to a draft PR which we will not be merging.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.h
```diff
@@ -39,5 +39,10 @@ namespace VehicleDynamics
         AZStd::vector<SteeringDynamicsData> m_steeringData;
         PidConfiguration m_steeringPid;
         PidConfiguration m_speedPid;
+
+        float maxSpeedImpulse = 0.0f;
```
adamdbrw: This does not look like a good default. The component in starting state would prevent vehicle from moving. 
Set this to whatever makes sense for a typical vehicle as a maximum. This should not prevent things working out of the box unless someone has an uncommon use case.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
adamdbrw: Also, do we need to limit steering impulse as well? If so, then adding maxCommandValue to PIDConfiguration instead would be better.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.h
adamdbrw: I understand that 0 means disable.

adamdbrw: If so, also add a comment next to this default.

adamdbrw: Would no limit provide a better experience for a user as a default than a limit set to some value, say 500?

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
pawelbudziszewski: I believe we can find (and test) a formula for this, but IMO manual tuning will be still useful. Maybe we should place settings like this in "deeper" parts of the config tree, something like `Advanced settings`.

As for the steering force limit: currently, it seems to be working fine without it, but there is a chance (or risk) that we'll be forced to add it. However, these limits may have different values, so we can't use just one value.

adamdbrw: If moved to PIDConfiguration, these would be two values, defaults could be overwritten in the SimplifiedDriveModel when constructing these. But for now we can leave it as it is with one value if the second one would not be used.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
pawelbudziszewski: Done. The new PR is on the way.

```
We need to have 3 fields for speed and 3 for steering, a nice layout of 3x2 would be good to present values.
```
I personally prefer to be able to see debug info from one PID at a time only.

```
That is, if you believe it would be useful for users in general. If it is only for us, then move it out to a draft PR which we will not be merging.
```
I don't believe we can quite get away from manual PID tuning when applying the controller to different vehicles (even if we manage to set some values automatically). And this output is useful in such cases. So I'd leave it, but move to the `Advanced settings` part of the tree.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.h
pawelbudziszewski: Added comment in the code.

You're right, when zero is set, the limit is disabled (it's mentioned in the pop-up message). I'm not sure, what is the best value, even for the kraken (not mentioning any general formula). However, when I think about it, the reasonable starting (default) value would be great, especially for our demo. And this applies to all settings around (mainly PIDs). The best, general solution would be to calculate these values automagically, but I'm not sure we have time/resources to develop this (and - what's probably more time-consuming - to test this) right now. But that's a topic for a different discussion.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
pawelbudziszewski: Good point. Ok, let's leave it.

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.cpp
```diff
@@ -134,6 +142,18 @@ namespace VehicleDynamics
                 continue;
             }
 
+            if (maxSpeedImpulse > 0.0)
+            {
+                if (pidCommand > maxSpeedImpulse)
+                {
```
adamdbrw: better:

```
if (maxSpeedImpulse > 0.0) 
{ // 0.0 means no limit will be applied
    pidCommand = AZStd::clamp(pidCommand, -maxSpeedImpulse, maxSpeedImpulse);
}
```

adamdbrw: (just apply formatting to the above

Code/Source/VehicleDynamics/DriveModels/SimplifiedDriveModel.h
```diff
@@ -39,5 +39,7 @@ namespace VehicleDynamics
         AZStd::vector<SteeringDynamicsData> m_steeringData;
         PidConfiguration m_steeringPid;
         PidConfiguration m_speedPid;
+
+        float maxSpeedImpulse = 0.0f; // Zero means, the imuplse limit is disabled
```
adamdbrw: Correct spelling and punctuation here

pawelbudziszewski:
Replaced by #242.

PR #240, opened 2022-09-28
Vehicle debug output

pawelbudziszewski:
Printing debug information for speed and steering (useful for PID tuning). Can be turned on/off using switches in the GUI.


adamdbrw:
There are additional changes in this PR with limits. Please only have debug changes. Also, I believe they could be moved to our Pid configuration class as well (like output limits) with just a tiny API change (form error to two arguments which then can be debugged and substracted to compute error)

michalpelka:
I would not merge it at this state - it is an important feature, but I think the proper debug output should be presented in ImGUI.

PR #242, opened 2022-09-29
Limit PID output

pawelbudziszewski:
More elegant and universal solution for limiting PID output (as #239), suggested by @adamdbrw. 

Added output limit to the PID object, with GUI control. The limit can be applied to all controllers using PID (vehicle steering and speed).

This PR replaces #239.


Code/Source/VehicleDynamics/DriveModels/PidConfiguration.h
```diff
@@ -37,6 +37,7 @@ namespace VehicleDynamics
         double m_iMax = 10.0; //!< maximal allowable integral term
         double m_iMin = -10.0; //!< minimal allowable integral term
         bool m_antiWindup = false; //!< prevents condition of integrator overflow in integral action
+        double m_outputLimit = 0.0; //!< prevents condition of integrator overflow in integral action
```
adamdbrw: Just a comment to fix here - it is copied from the line above.

adamdbrw (commented):
One minor fix remains - a comment

adamdbrw (approved):

PR #247, opened 2022-09-30
Added routine to access components without specifying if wrapped

michalpelka:
A lot of components in this Gem are created wrapped in GenericComponentWrapper. Those need to be accessed differently in game mode and editor. `GetGameOrEditorComponent` was added to help with that.

### DIFF
--- a/Code/Source/ROS2GemUtilities.h
+++ b/Code/Source/ROS2GemUtilities.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include "AzCore/Component/ComponentBus.h"
+#include "AzCore/Component/Entity.h"
+#include "AzToolsFramework/ToolsComponents/GenericComponentWrapper.h"
+
+namespace ROS2
+{
+    namespace Utils
+    {
+        /// Create component for a given entity in safe way.
+        /// \param entityId entity that will own component
+        /// \param componentType Uuid of component to create
+        /// \return created componentId, if it fails, it returns invalid id
+        AZ::ComponentId CreateComponent(const AZ::EntityId entityId, const AZ::Uuid componentType);
+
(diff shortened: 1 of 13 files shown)


Code/Source/Camera/CameraSensor.cpp
```diff
@@ -132,7 +132,7 @@ namespace ROS2
         AZ::Transform inverse = cameraPose.GetInverse();
         m_view->SetWorldToViewMatrix(AZ::Matrix4x4::CreateFromQuaternionAndTranslation(inverse.GetRotation(), inverse.GetTranslation()));
 
-        size_t userId = AZ::Render::FrameCaptureRequests::s_InvalidFrameCaptureId;
```
adamdbrw: Probably should not be here? It seems that this is an unrelated change.

Code/Source/ROS2GemUtilities.h
```diff
@@ -0,0 +1,46 @@
...
+        /// Retrieve component from entity given by a pointer. It is a way to get game components and wrapped components.
+        /// We should use that that we are not sure if we access e.g. ROS2FrameComponent in game mode or from Editor
+        /// \param entity pointer to entity e.g. with GetEntity()
+        /// \return pointer to component with type T
+
+        template<class ComponentType>
+        ComponentType* GetGameOrEditorComponent(const AZ::Entity* entity)
+        {
+            AZ_Assert(entity, "Called with empty entity");
+            ComponentType* component = entity->FindComponent<ComponentType>();
```
adamdbrw, line 33: Is there reasoning why we are looking in this order?

adamdbrw: I was wondering whether we could check whether we are in Game mode or Editor mode and retrieve based on that. Would that work?

adamdbrw (commented):
Important fix. Could be accepted as it is, but please take a look at the comments first.

Code/Source/ROS2GemUtilities.h
michalpelka: Yes, the logic is that game mode is more time critical.

Code/Source/Camera/CameraSensor.cpp
michalpelka: Done

adamdbrw (approved):

PR #266, opened 2022-10-10
Spawn point quaternion fix

pawellech1:
Signed-off-by: Paweł Lech <pawel.lech@robotec.ai>

### DIFF
--- a/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -13,6 +13,7 @@
 #include <AzCore/Serialization/EditContext.h>
 #include <AzFramework/Components/TransformComponent.h>
 #include <AzToolsFramework/Entity/EditorEntityHelpers.h>
+#include <ROS2/Utilities/ROS2Conversions.h>
 
 namespace ROS2
 {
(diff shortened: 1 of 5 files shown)


Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -213,10 +214,7 @@ namespace ROS2
             pose.position.x = info.pose.GetTranslation().GetX();
             pose.position.y = info.pose.GetTranslation().GetY();
             pose.position.z = info.pose.GetTranslation().GetZ();
-            pose.orientation.x = info.pose.GetRotation().GetX();
-            pose.orientation.y = info.pose.GetRotation().GetY();
-            pose.orientation.z = info.pose.GetRotation().GetZ();
-            pose.orientation.w = info.pose.GetRotation().GetW();
+            pose.orientation = ROS2Conversions::ToROS2Quaternion(info.pose.GetRotation());
```
piotr-zyskowski-rai: It would be nice to have `ToROS2Pose` function. I might not remember correctly but conversion from AZ::Vector3 to ROS vector3 is also available so it could be used there as well

piotr-zyskowski-rai (approved):
LGTM

PR #271, opened 2022-10-11
Getting ROS2 Frame from any ancestor instead of parent

### LINKED ISSUE #270: Extend looking for frames propagation
**Is your feature request related to a problem? Please describe.**
While composing prefabs in prefabs, a `FrameComponent` from an inside prefab stops its "parent" discovery on the dummy entity ( (which is added automatically every time a prefab is added to the hierarchy). This results in, ie sensor not inheriting the parent namespace by default and needs to be set up manually.

**Describe the solution you'd like**
Look for parent `FrameComponent` until the root of the o3de tree.

adamdbrw:
Fixes #270

I tested once that it does not break navigation.
However I am uncertain how you would like to use it, so please test


pijaro (approved):
lgtm, works

PR #283, opened 2022-10-13
Added Ignore layer param for lidar. Removed ignored entity

pijaro:


adamdbrw (approved):

adamdbrw (approved):
As a first step before doing it with the actual masks

PR #295, opened 2022-10-26
Urdf creation in passes

### LINKED ISSUE #233: URDF importer creates multiple entities with the same name
**Bug description**
Let us assume that we have a URDF with cycles. 
You can use URDF to describe parallel structures like:
```` 
          A
         /   \
       B      C 
        \    /
          D
```` 
The URDF importer will create multiple entities `D`           
**To Reproduce**
1. Get some URDF with parallel structure
2. Link that is referred to in the different branches is created more than one time

**Expected behavior**
Correctly construct a parallel structure.

**Screenshots / videos**
N/A

**Test environment**
 - OS: Ubuntu 22.04
 - ROS: Humble

michalpelka:
Changed way in which URDF is parsed:
 - entities are created without hierarchy at first
 - entities transformations are set independent of joint
 - Widgets contain log of created entities and joints
 - Fixed crash in ColliderMaker, that was caused by a non-existing mesh collider
 
Fixes #233


adamdbrw (approved):

adamdbrw:
Added some language fixes

PR #298, opened 2022-11-01
Prevent non-linux builds of the ROSGem

spham-amzn:
Add PALification code to restrict this gem to only build on Linux (for now)

### DIFF
--- a/Code/CMakeLists.txt
+++ b/Code/CMakeLists.txt
@@ -3,10 +3,18 @@
 #
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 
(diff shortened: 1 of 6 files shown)


lemonade-dm (approved):

adamdbrw (approved):
LGTM, thank you!

adamdbrw:
Did not build on our CI machine, but it is an issue with the machine. I tested locally.

spham-amzn:
> Did not build on our CI machine, but it is an issue with the machine. I tested locally.

Its because the machine didn't have ROS sourced. I also changed it so configure fails if it cannot detect that ROS2 was installed or sourced

Issue #196, opened 2022-09-16
Editor stops working on attempt to reaload URDF when prefab is edited


michalpelka:
**Bug description**
Editor stops when previously URDF is opened and edited (as prefab) and user tries to reload URDF
**To Reproduce**
1. Go to 'Robot importer'
2. Click on 'import'
3. Double click on created prefab to open it. 
4. Go to 'Robot importer'
5. Click on 'import'
6 Editor stopped.

**Expected behavior**
It would be good to disable robot import when any prefab is edited
or
Try to load URDF, but check if this prefab is safe to remove

**Screenshots / videos**
If applicable, add screenshots/videos to help explain your problem.

**Test environment**
 - OS: Ubuntu 22.04
 - ROS: Humble

michalpelka:
It can be closed, it was fixed with #300.

PR #300, opened 2022-11-03
Cyclical prefab import blocked

pawellech1:
solves [issue](https://github.com/RobotecAI/o3de-ros2-gem/issues/196)

### DIFF
--- a/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -83,6 +83,13 @@ namespace ROS2
                     ReportError("User cancelled");
                     return;
                 }
+
+                if (IsDependencyCyclical(prefabName.Filename()))
+                {
+                    ReportError("Cyclical dependency detected.\nPlease choose another prefab to import or exit the focus mode.");
+                    return;
+                }
+
                 m_robotImporter.ParseURDFAndStartLoadingAssets({ urdfPath.value(), prefabPath->c_str() });
 
                 // Disable the button until the import is complete to prevent the user from clicking it again
@@ -92,6 +99,18 @@ namespace ROS2
             });
 
         setLayout(mainLayout);
+    }
+
+    bool RobotImporterWidget::IsDependencyCyclical(const AZ::IO::PathView& importedPrefabFilename)
+    {
+        AzFramework::EntityContextId context_id;
+        EBUS_EVENT_RESULT(context_id, AzFramework::EntityIdContextQueryBus, GetOwningContextId);
+
+        auto focus_interface = AZ::Interface<AzToolsFramework::Prefab::PrefabFocusInterface>::Get();
+        auto focus_prefab_instance = focus_interface->GetFocusedPrefabInstance(context_id);
+        auto focus_prefab_filename = focus_prefab_instance.value().get().GetTemplateSourcePath().Filename();
+
+        return focus_prefab_filename == importedPrefabFilename;
     }
 
     void RobotImporterWidget::ReportError(const AZStd::string& errorMessage)
--- a/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -84,9 +84,8 @@ namespace ROS2
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -94,6 +101,22 @@ namespace ROS2
...
+
+        auto focus_interface = AZ::Interface<AzToolsFramework::Prefab::PrefabFocusInterface>::Get();
+        auto focus_prefab_instance = focus_interface->GetFocusedPrefabInstance(context_id);
+        auto focus_prefab_filename = focus_prefab_instance.value().get().GetTemplateSourcePath().Filename();
+
+        if (focus_prefab_filename == importedPrefabFilename)
+        {
+            return true;
+        }
+        return false;
```
pijaro:
```suggestion
        return focus_prefab_filename == importedPrefabFilename;
```

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -94,6 +101,18 @@ namespace ROS2
         setLayout(mainLayout);
     }
 
+    bool RobotImporterWidget::IsDependencyCyclical(const AZ::IO::PathView& importedPrefabFilename)
+    {
+        AzFramework::EntityContextId context_id;
+        EBUS_EVENT_RESULT(context_id, AzFramework::EntityIdContextQueryBus, GetOwningContextId);
+
+        auto focus_interface = AZ::Interface<AzToolsFramework::Prefab::PrefabFocusInterface>::Get();
```
adamdbrw, line 108: Is this always non-null?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -94,6 +101,18 @@ namespace ROS2
...
+        auto focus_prefab_instance = focus_interface->GetFocusedPrefabInstance(context_id);
+        auto focus_prefab_filename = focus_prefab_instance.value().get().GetTemplateSourcePath().Filename();
```
adamdbrw, line 124: What if the instance is null?

Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -83,6 +83,13 @@ namespace ROS2
                     ReportError("User cancelled");
                     return;
                 }
+
+                if (IsDependencyCyclical(prefabName.Filename()))
+                {
+                    ReportError("Cyclical dependency detected.\nPlease choose another prefab to import or exit the focus mode.");
```
pijaro: I think we should use more user-friendly descriptions about what is wrong and what to do.

Maybe something like:
```suggestion
                    ReportError("Selected URDF model is currently being edited. Exit prefab edit mode and try again.");
```
?

pawellech1: I mentioned cyclical dependency basing on the O3DE notification which shows up when you try to manually instantiate prefab inside itself:
image

michalpelka: It is ok that is a cyclical dependency, but we should prompt users with a more user-friendly message.

adamdbrw:
@pawellech1 what is the status here? There are unresolved conversations

Code/Source/RobotImporter/RobotImporterWidget.cpp
pawellech1: should the import be aborted if any of these is null?

pawellech1: Since it is actually wrapped reference of instance I don't think it might be null but I checked if std::optional is std::nullopt before using the instance.

Code/Source/RobotImporter/RobotImporterWidget.cpp
pawellech1: I changed the error text.

adamdbrw:
@pawellech1 what is the status? Could you resolve conversations which are resolved?

pawellech1:
@adamdbrw I am not sure if I can resolve this one https://github.com/RobotecAI/o3de-ros2-gem/pull/300#pullrequestreview-1166857883

adamdbrw (approved):

PR #320, opened 2022-11-15
RGBD Camera

michalpelka:


pawellech1:
Did you consider creating CameraDepthSensor as a separate component from CameraRGBSensor?

michalpelka:
I would not decouple it into two components. Most use cases call for an RGBD camera with the same viewport. One can set it up manually if one needs to simulate a more realistic RGBD camera with some parallax.

michalpelka:
Tested against:
o3de : cb09e08b75e66c509e80e6f17a9981928b8d3a59
RosConDemo: 6922187247b0b23e4bb91fd1fba09060b42ab76c

Code/Source/ROS2SystemComponent.cpp
```diff
@@ -132,4 +147,16 @@ namespace ROS2
...
     }
+
+    void ROS2SystemComponent::LoadPassTemplateMappings()
+    {
+        AZ_Printf("ROS2CameraSensorComponent", "LoadPassTemplateMappings\n");
+        auto* passSystem = AZ::RPI::PassSystemInterface::Get();
+        AZ_Assert(passSystem, "PassSystemInterface is null");
+
+        const char* passTemplatesFile = "Passes/ROSPassTemplates.azasset";
+        bool isOk = passSystem->LoadPassTemplateMappings(passTemplatesFile);
```
michalpelka:
```suggestion
        [[maybe_unused]] bool isOk = passSystem->LoadPassTemplateMappings(passTemplatesFile);
```

Code/Source/ROS2SystemComponent.cpp
```diff
@@ -86,6 +90,17 @@ namespace ROS2
         m_staticTFBroadcaster = AZStd::make_unique<tf2_ros::StaticTransformBroadcaster>(m_ros2Node);
         m_dynamicTFBroadcaster = AZStd::make_unique<tf2_ros::TransformBroadcaster>(m_ros2Node);
 
+        AZ_Printf("FOOROS2SystemComponent", "Hello!\n");
```
adamdbrw: Remove this printf please

Code/Source/ROS2SystemComponent.cpp
```diff
@@ -86,6 +90,17 @@ namespace ROS2
...
+        auto* passSystem = AZ::RPI::PassSystemInterface::Get();
```
adamdbrw, line 93: Is this the right place for this? I understand this is something required for the camera sensor.

Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -44,11 +44,16 @@ namespace ROS2
...
+        bool m_colorCamera = true;
+        bool m_depthCamera = true;
 
         void FrequencyTick() override;
-
         std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::Image>> m_imagePublisher;
+        std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::Image>> m_depthPublisher;
         std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::CameraInfo>> m_cameraInfoPublisher;
-        std::optional<CameraSensor> m_cameraSensor;
+        std::optional<CameraSensor> m_cameraSensorColor;
```
adamdbrw: How about an approach similar to Lidar? 
Implementation of data acquisition itself could be delegated to another class (akin to LidarRaycaster). There is also some common code between cases. Handling of Ifs could be internal to implementations.

Code/Source/ROS2SystemComponent.cpp
michalpelka: Sorry for debugging - I forgot to remove it. 
Yes, we need to add it here please refer other gems with custom passes: https://github.com/o3de/o3de/blob/development/Gems/Terrain/Code/Source/Components/TerrainSystemComponent.cpp#L70

Code/Source/Camera/ROS2CameraSensorComponent.h
michalpelka: Decoupled and use virtualization to remove duplicated code.

adamdbrw (approved):

PR #321, opened 2022-11-18
Camera editor component

michalpelka:


michalpelka:
Tested against:
o3de : cb09e08b75e66c509e80e6f17a9981928b8d3a59
RosConDemo: 6922187247b0b23e4bb91fd1fba09060b42ab76c

Code/Source/Camera/ROS2CameraSensorEditorComponent.cpp
```diff
@@ -0,0 +1,145 @@
+#include "ROS2CameraSensorEditorComponent.h"
+#include "ROS2CameraSensorComponent.h"
+namespace ROS2
+{
+
+    ROS2CameraSensorEditorComponent::ROS2CameraSensorEditorComponent()
+    {
+        m_sensorConfiguration.m_frequency = 10;
+        m_sensorConfiguration.m_publishersConfigurations.insert(
+            Internal::MakeTopicConfigurationPair("camera_image_color", Internal::kImageMessageType, Internal::kColorImageConfig));
```
pawelbudziszewski: I've got a compilation error:
`error: use of undeclared identifier 'Internal'`

Code/Source/Camera/ROS2CameraSensorEditorComponent.cpp
```diff
@@ -0,0 +1,145 @@
...
+        m_sensorConfiguration.m_publishersConfigurations.insert(
+            Internal::MakeTopicConfigurationPair("camera_image_depth", Internal::kImageMessageType, Internal::kDepthImageConfig));
+        m_sensorConfiguration.m_publishersConfigurations.insert(
+            Internal::MakeTopicConfigurationPair("camera_info", Internal::kCameraInfoMessageType, Internal::kInfoConfig));
+    }
```
pawelbudziszewski, line 25: ...and for all other `Internal`-s in this part

pawelbudziszewski (changes requested):

pawelbudziszewski:
Is there any particular reason why this camera is created along the Z-axis? The O3DE camera is always looking along the Y-axis. It might be helpful while setting the ROS camera if it was rotated in the same way as the O3DE cam.

Screenshot from 2022-11-23 13-24-58

michalpelka:
Yes, there is a very particular reason and it is  - it is an image processing coordinate system. O3de uses more entertainment industry standards for camera coordinate.

Here is a quote from [REP-103:](https://www.ros.org/reps/rep-0103.html#id22)

```
In the case of cameras, there is often a second frame defined with a "_optical" suffix. This uses a slightly different convention:

    z forward
    x right
    y down
```    
I will defend this coordinate system - as an essential thing here. What is more, there is no easy way to implement what you mentioned @pawelbudziszewski  - note that the camera needs a **Ros2SensorComponent** - and the **Ros2SensorComponent** is not aware if it is part of the camera or not.
Implementing such a "quirk" would result in a number of unnecessary discrepancies, that would render the whole component unusable.  That tool is intended for ROS2 users - that community would appreciate that we are using well-established standards.

Code/Source/Camera/ROS2CameraSensorEditorComponent.cpp
```diff
@@ -0,0 +1,154 @@
...
+        gameEntity->CreateComponent<ROS2::ROS2CameraSensorComponent>(
+            m_sensorConfiguration, m_VerticalFieldOfViewDeg, m_width, m_height, m_colorCamera, m_depthCamera);
+    }
+
+    void ROS2CameraSensorEditorComponent::DisplayEntityViewport(
+        const AzFramework::ViewportInfo& viewportInfo, AzFramework::DebugDisplayRequests& debugDisplay)
+    {
+        AZ_UNUSED(viewportInfo);
+        AZ::u32 stateBefore = debugDisplay.GetState();
+        AZ::Transform transform = GetEntity()->GetTransform()->GetWorldTM();
```
pawelbudziszewski, line 106: Compilation error:

`o3de-ros2-gem/Code/Source/Camera/ROS2CameraSensorEditorComponent.cpp:96:62: error: member access into incomplete type 'AZ::TransformInterface'`

Including `AzCore/Component/TransformBus.h` resolves the problem

pawelbudziszewski (changes requested):

Code/Source/Camera/ROS2CameraSensorEditorComponent.cpp
michalpelka: Done

pawelbudziszewski (approved):
Works fine now. I have no further comments.

Tested:
- All parameters seem to work correctly, in particular:
- FOV seems to be set correctly (at least images look similar to the O3DE cam image)
- Depth camera values seem reasonable
- All data can be accessed via ROS topics

As for the camera rotation: I accept the explanation.

adamdbrw:
This is a very useful PR. I would love this to connect with Visualise() of ROS2SensorComponent.

Visualise could be different for Editor and Game mode components (for example, lidar as frustum vs as points). Perhaps there should be a ROS2SensorEditorComponent?

michalpelka:
This PR will be moved to https://github.com/o3de/o3de-extras/

PR #297, opened 2022-10-27
Urdf import wizard

michalpelka:
Significant improvements to URDF importer:
 - new feel with Wizard :mage: 
 - a more robust way to create entities
 - little more extensive test coverage
 - checksum-based resolvement of URDF meshes - potential fix for o3de/o3de-ros2-gem#282 , it covers some of subtasks of o3de/o3de-extras#148

PR #328, opened 2022-11-23
Support for loading XACRO

michalpelka:

### DIFF
--- a/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
+++ b/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
@@ -10,6 +10,7 @@
 #include <AzCore/Math/MathStringConversions.h>
 #include <QHeaderView>
 #include <QVBoxLayout>
+
 namespace ROS2
 {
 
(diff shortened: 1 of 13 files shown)


pawelbudziszewski (commented):
Works fine with xacro files.

Some minor comments:
- In the wizard descriptions, I'd change `URDF` to `URDF/XACRO`(or something like this)
- `This wizard allows you to build robot simulation [...]` text should be updated just like in the #297

adamdbrw:
Looks good - just please add a task to handle parameters and some comments in the code - perhaps it would be best to extract xacro loading to a subroutine

pawelbudziszewski (commented):
Works as expected, descriptions are fine. I have just a minor comment: 
Screenshot from 2022-12-09 11-22-45
In case of missing assets, would it be possible to write in the last column something like `Asset not found!`? The red color suggests, that there is a problem with such asset, but writing it explicitly would make it more obvious.

michalpelka:
Was replaced with : https://github.com/o3de/o3de-extras/pull/90

Issue #311, opened 2022-11-09
 Standards and styles / Miscellaneous / Part II


adamdbrw:
Apply the following guidelines to the code base:
- [x] [RTTI](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#rtti)
- [x] [Casting](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#casting)
- [x] [Use of const](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#use-of-const)
- [x] [Don’t leave disabled code in the source](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#dont-leave-disabled-code-in-the-source)

michalpelka:
There is still some unused code :
https://github.com/RobotecAI/o3de-ros2-gem/blob/008c13dccfef7df9cc9745c2df49a35b173e1b90/Assets/Scripts/SimpleTwistControl.lua

Issue #310, opened 2022-11-09
Standards and styles / Miscellaneous / Part I


adamdbrw:
Apply the following guidelines to the code base:
- [x] [Use AZ or AZStd](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#use-az-or-azstd)
- [x] [Use of auto](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#use-of-auto)
- [x] [Use of nullptr](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#use-of-nullptr)
- [x] [C++ Exceptions](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#c-exceptions)

forhalle:
Thanks for your work on this @michalpelka! I'm re-opening, however, because we would like to include at least one SIG reviewer who was involved in this project (i.e. Derric, Steve, Alex, Tom, etc.) on each task to review the standards. 

Please feel free to continue work on the tasks on this board, but please do not set them to "Done"... We will let the SIG reviewers do that. Instead, feel free to leave comments in the tasks as to the work you have performed.

PR #329, opened 2022-11-25
Standards and styles : std usage, auto, nullptr, rtti

michalpelka:


michalpelka:
Here I went through the code base and covered:
#310 and RTTI task from #311
# Use AZ or AZStd
Changed some std to AZStd, please note that some portion of the codebase (also in ROS.Static) links a lot from the standard library through `rclcpp`.
# Use of auto
Adjusted usage of `auto` to match [guideline](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#use-of-auto)
# Use of nullptr
No problem here
# Exceptions and RTTI
I disabled bort RTTI and exception from ROS2.Static/ROS2 targets. 
Note that still there is need for RTTI and exceptions in ROS2.Editor. 
It is caused mainly by URDF parse library. It could be solved by incorporating that library into codebase - disabling `std::runtime_error`

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -32,13 +30,19 @@ namespace ROS2
...
+            { AZ::RHI::Format::R32_FLOAT, "32FC1" },
+        };
 
+        AZStd::unordered_map<AZ::RHI::Format, int> BitDepth{
+            { AZ::RHI::Format::R8G8B8A8_UNORM, 4 * sizeof(uint8_t) },
+            { AZ::RHI::Format::R16G16B16A16_UNORM, 4 * sizeof(uint16_t) },
+            { AZ::RHI::Format::R32G32B32A32_FLOAT, 4 * sizeof(float) }, // Unsuported by RVIZ2
+            { AZ::RHI::Format::R8_UNORM, sizeof(uint8_t) },
+            { AZ::RHI::Format::R16_UNORM, sizeof(uint16_t) },
+            { AZ::RHI::Format::R32_FLOAT, sizeof(float) },
```
adamdbrw, line 49: This change block seems to be unrelated to style changes

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -188,10 +192,10 @@ namespace ROS2
                 AZ_Assert(Internal::FormatMappings.contains(format), "Unknown format in result %u", static_cast<uint32_t>(format));
                 sensor_msgs::msg::Image message;
                 message.encoding = Internal::FormatMappings.at(format);
+
```
adamdbrw: unnecessary newline

Code/Source/Camera/CameraSensor.cpp
```diff
@@ -188,10 +192,10 @@ namespace ROS2
...
                 message.width = descriptor.m_size.m_width;
                 message.height = descriptor.m_size.m_height;
-                message.step = message.width * sensor_msgs::image_encodings::bitDepth(message.encoding) / 8 *
-                    sensor_msgs::image_encodings::numChannels(message.encoding);
+                message.step = message.width * Internal::BitDepth.at(format);
```
adamdbrw, line 201: unrelated?

Code/Source/Camera/CameraSensor.cpp
michalpelka: I disagree - it is related to the elimination of the usage of exception, what is mentioned in [style](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#c-exceptions). With this change I do not need to include `sensor_msgs/image_encodings.hpp` that uses exception in header's code

Code/Source/Camera/CameraSensor.cpp
michalpelka: I disagree - it is related to the elimination of the usage of exception, what is mentioned in [style](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#c-exceptions). With this change I do not need to include `sensor_msgs/image_encodings.hpp` that uses exception in header's code

adamdbrw (approved):

PR #330, opened 2022-11-25
Standards and styles : const

michalpelka:
Addresses only task #311

# Use of const
I've tried to add everywhere possible `const`


Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -1,4 +1,4 @@
-/*
+/*const
```
adamdbrw: Does look like a mistake

adamdbrw (approved):

PR #226, opened 2022-09-22
Design docs for manipulation

j-rivero:
Fixes o3de/o3de-extras#151 

Draft documents about the implementation of the manipulation feature for the o3de-ros2-gem. Mostly inspired by o3de/o3de-ros2-gem#162.

**List of files:**
 * design.md: main intro file
 * analysis.md: detailed description about what we understand as manipulation, how does it applies to the o3de-ros2-gem
 * architecture.md: (TODO) class diagram for the implementation
 * features.md: list of features desired for the manipulation feature
 * manipulation.svg: informal diagram of data flow and interactions between o3de and ROS 2 for the different aspects of the manipulation.

I left the`architecture.md` to be done since I have no previous knowledge of the gem/o3de internals. Any guidance or help is very welcome. I left the PR as Draft by now but I can happy answer to any feedback on the current content.

PR #358, opened 2022-12-08
Manipulation implementation roadmap

mabelzhang:
Targeting #226

Add implementation roadmap for the simplified design diagram in https://github.com/RobotecAI/o3de-ros2-gem/pull/226#pullrequestreview-1178605453

@j-rivero please review.


docs/design/manipulation/analysis.md
```diff
@@ -124,6 +124,9 @@ To implement these set of actions, the O3DE simulator should have at least:
 
  * Provide data from the simulated sensors: not in the scope of this document
 
+For an initial implementation, a lighter version of the diagram can be an intermediate goal (simplified by danielemorra98):
+
+![Manipulation light](https://user-images.githubusercontent.com/48955695/201618734-d8998631-1472-4326-822e-c5ddd66db903.svg)
```
j-rivero: The image has the background transparent which makes it difficult to read under dark themes configurations. It would be great if we can add a white background to it.

j-rivero: white-background

I'll change it.

docs/design/manipulation/roadmap.md
```diff
@@ -0,0 +1,58 @@
...
+- Determine ROS topic and type (e.g.
+  [`geometry_msgs/PoseStamped`](https://docs.ros2.org/latest/api/geometry_msgs/msg/PoseStamped.html)),
+  to publish object pose in simulator to ROS.
+
+## Motion planning
+
+- Scripting
+  - Calculate end-effector goal pose based on object pose in simulator.
+    Publish end-effector goal pose for MoveIt.
+- Create `ROS2JointStatePubComponent` to obtain joints information from
```
j-rivero, line 35: If I understand correctly, when using `ros2_control` the `/joint_state` topic is being populated by the [joint_state_broadcaster](https://control.ros.org/master/doc/ros2_controllers/joint_state_broadcaster/doc/userdoc.html). If this is correct, this could a simplified version of the approach. We can mention this in the same way that we did for  https://github.com/RobotecAI/o3de-ros2-gem/pull/358/files#diff-7eacaf4f964ca57310bf9329bd15ce9de0ec8a17060caf87ce9a239ca99b2c55R46-R55.

docs/design/manipulation/analysis.md
mabelzhang: Are you saying you'll change it in your PR?
Either way, I changed it here too so that this PR goes in with a good version a64574464f85615f270f240fa0b979269db8d725

Currently the image is referenced twice - once in analysis.md, again in roadmap.md so that the picture can be on the same page as the roadmap for reference. It's not great to have duplicates, as we might change one and forget to change the other (I changed it in both places just now). Should we have it only in roadmap.md, and remove the light version in analysis.md?

docs/design/manipulation/roadmap.md
mabelzhang: bdfd7af

docs/design/manipulation/analysis.md
j-rivero: How about https://github.com/mabelzhang/o3de-ros2-gem/pull/1/files so we include the file in the repo and reference it from here?

> Should we have it only in roadmap.md, and remove the light version in analysis.md?

I think is fine to leave it two places, not a big deal.

j-rivero (approved):

j-rivero (commented):
ready to go as soon as we solve the small issue with svg images.

docs/design/manipulation/analysis.md
mabelzhang: 1d84eff

